PR #1, opened 2023-05-31
feat: add first slog version

katcipis:
Just adding our current slog. After we tag it we can start improving. Already have better ideas on how to handle adding fields and propagating the log context (the current way doesn't scale well and involve custom handlers). So we should have some improvements on the API coming soon.

The slog library is a thin wrapper around the experimental standard structured log that will be added on Go soon (it was already accepted). We can shift to the standard lib implementation with no impact on our code + there is some extra functionality that a more general/standard library won't do, ours is more specific in assuming things like tracing and config loading from env vars.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,4 @@
-# golibs
-Birdie's standard library
+# Go Libs
+
+Birdie's standard library. This repository works as a monorepo of different Go modules
+that we use on our services and tools here at Birdie.
--- a/slog/go.mod
+++ b/slog/go.mod
@@ -1,0 +1,8 @@
+module github.com/birdie-ai/golibs/slog
+
+go 1.20
+
+require (
+	github.com/google/uuid v1.3.0
+	golang.org/x/exp v0.0.0-20230522175609-2e198f4a06a1
+)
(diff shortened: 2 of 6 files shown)


guilherme-santos (approved):
LGTM, but I'd like to discuss after, if `tracing.Instrument` should be inside of slog package, I don't see log-related stuff there

katcipis:
> LGTM, but I'd like to discuss after, if `tracing.Instrument` should be inside of slog package, I don't see log-related stuff there

It could be a separate tracing only package, it only references the slog tracing key, but even that reference should be refactored away in a new design.

katcipis:
> > LGTM, but I'd like to discuss after, if `tracing.Instrument` should be inside of slog package, I don't see log-related stuff there
> 
> It could be a separate tracing only package, it only references the slog tracing key, but even that reference should be refactored away in a new design.

Just to re-iterate...this is NOT a good/final design :rofl:. You can see it as a draft with a lot of bad ideas. My goal is to have a first tag I can reference and already improve it.

guilherme-santos:
> > > LGTM, but I'd like to discuss after, if `tracing.Instrument` should be inside of slog package, I don't see log-related stuff there
> > 
> > 
> > It could be a separate tracing only package, it only references the slog tracing key, but even that reference should be refactored away in a new design.
> 
> Just to re-iterate...this is NOT a good/final design 🤣. You can see it as a draft with a lot of bad ideas. My goal is to have a first tag I can reference and already improve it.

Yeah, it was clear to me... That's why I approved but left the comment just that we can have it in mind! :)

PR #2, opened 2023-05-31
chore: add basic CI

katcipis:
Add basic lint/testing on the monorepo. Didn't imposed a Makefile for now.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,7 +3,7 @@ on: [push]
 
 jobs:
   lint:
-    name: Test
+    name: Lint
 
     runs-on: ubuntu-22.04
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,18 +1,41 @@
 name: CI
 on: [push]
 
-env:
-  modules:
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,54 @@
...
+jobs:
+  lint:
+    name: Lint
+
+    runs-on: ubuntu-22.04
+
+    strategy:
+      matrix:
+        module:
+          - slog
```
katcipis, line 13: Did not find a nice way to avoid duplicating this module list, there is no proper variables in GHA, only env vars (which are always just a plain string).

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+  test:
+    name: Test
+
+    runs-on: ubuntu-22.04
+
+    strategy:
+      matrix:
+        go:
+          - "1.20"
```
katcipis, line 39: since we may use different Go versions around different services at least in a time window, we can test with different Go versions

guilherme-santos (approved):

guilherme-santos (approved):

PR #3, opened 2023-06-01
feat: add new context save/load for loggers

katcipis:
The idea is to simplify how we add logging context. The current approach would not scale well for dynamic fields + required custom handlers. The new approach is simpler, just add any logger you want to a context and then retrieve that logger from that context, that is it.

I already refactored the tracing package outside slog, since it is unrelated and builds on top of it.

Usually I would split this, but since the library is still on its infancy I thought it would be easier to see these changes together (they are different modules, so usually you need to update one module first and then update the other one).


tracing/tracing_test.go
```diff
@@ -0,0 +1,25 @@
...
+		got = slog.FromCtx(req.Context())
+	}))
+
+	req := httptest.NewRequest(http.MethodGet, "/", nil)
+	res := httptest.NewRecorder()
+	handler.ServeHTTP(res, req)
+
+	if got == nil {
+		t.Fatal("got nil logger")
+	}
```
katcipis, line 24: The tests are not great right now, I wanted to test if the trace_id field is there, test the header integration, etc. But baby steps =P.

slog/slog.go
```diff
@@ -170,43 +160,29 @@ func With(args ...any) *Logger {
...
-}
-
-func getValue(ctx context.Context, key any) (string, bool) {
-	val := ctx.Value(key)
-	if val == nil {
-		return "", false
-	}
-	str, ok := val.(string)
+// FromCtx gets the [Logger] associated with the given context. An empty [Logger] is
+// returned if the context has no [Logger] associated with it.
```
guilherme-santos, line 164: We don't really return an empty log but the default one from slog, someone could update before for example. My point is just fix the comment, saying that if empty slog.Default() will be returned

slog/slog.go
```diff
@@ -170,43 +160,29 @@ func With(args ...any) *Logger {
...
 	}
-	return str, true
+	return log
 }
 
-func (t *tracedHandler) WithAttrs(attrs []slog.Attr) slog.Handler {
-	return t.handler.WithAttrs(attrs)
+// CtxWithLogger creates a new [context.Context] with the given [Logger] associated with it.
+// Call [FromCtx] to retrieve the [Logger].
+func CtxWithLogger(ctx context.Context, log *slog.Logger) context.Context {
```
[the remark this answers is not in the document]
guilherme-santos: And maybe keep both the name the same like `slog.FromContext(ctx)`

tracing/tracing.go
```diff
@@ -2,16 +2,16 @@
...
 	"github.com/google/uuid"
 )
 
-// Instrument will instrument the given handler by adding tracing context on the
-// request context.
-func Instrument(h http.Handler) http.Handler {
+// InstrumentHTTP will instrument the given [http.handler] by adding a slog.Logger on the request context.
+// The logger will have `trace_id` added to it.
+// Use slog.FromCtx(ctx) to retrieve the logger.
+func InstrumentHTTP(h http.Handler) http.Handler {
```
guilherme-santos, line 14: Also, not really happy about this either, I'm wondering if this is tracing, or this is HTTP middleware (short answer will be, both 😂)

What do you think of starting a `http` package, this would be the middleware `http.Instrument(h)`, I think about the Org-ID for example, the implementation is exactly the same, would we call organization package? or `http.Auth()` maybe?

I think we'll at some point have a tracing package, but I feel here is not really tracing is just reading a header (like org-id will), what do you think?

tracing/tracing_test.go
guilherme-santos: Yep, good points.

slog/slog.go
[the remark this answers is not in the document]
guilherme-santos: I read it as a sentence... Give me the slog from the context. I also have the feeling is quite common to find FromContext, but I can be wrong

tracing/tracing.go
guilherme-santos: I think I was thinking something really different, but thinking better, maybe it's good the way that you did.

I was expecting to have all HTTP related stuff together, but as we're splitting different libs, then HTTP lib would define stuff that are tracing related (like the field name and header), which would be weird, if they were all in the same library I think would be ok.

Let's move on, like that, I think can be good!

another minor comment, is if Instrument is the best wording, we can move forward as I don't really have other options in my mind.

tracing/tracing.go
```diff
@@ -24,7 +24,10 @@ func Instrument(h http.Handler) http.Handler {
 			slog.Debug("header absent, generated UUID", "trace_id", traceid)
 		}
 
-		req = req.WithContext(context.WithValue(req.Context(), slog.TraceID, traceid))
+		log := slog.With("trace_id", traceid)
```
guilherme-santos, line 27: Maybe I'm missing something, but should we use the log from the ctx, enrich it and then set it back, something like:

```go
ctx := req.Context()
log := slog.FromContext().With("trace_id", traceid)
ctx := slog.WithContext(ctx, log)
req = req.WithContext(ctx)
```

tracing/tracing.go
katcipis: Im very torn on these different design decisions, not strong feelings so far. If we stick with tracing I think tracing.InstrumentHTTP could be enough since it should have only a single tracing related function (but I specified the procotol because we could have other forms of instrumentation that are tracing related but are not HTTP...maybe actual tracing =P).

Regarding the OrgID, I would also place it on the tracing.InstrumentHTTP..I just want something that solves all our problems regarding basic traced data on HTTP..but maybe that is bad...and in that case we would need a different package like organization.InstrumentHTTP...but not sure we do... both designs are valid and have tradeoffs and Im on the fence on this.

katcipis: Im very torn on these different design decisions, not strong feelings so far. If we stick with tracing I think tracing.InstrumentHTTP could be enough since it should have only a single tracing related function (but I specified the procotol because we could have other forms of instrumentation that are tracing related but are not HTTP...maybe actual tracing =P).

Regarding the OrgID, I would also place it on the tracing.InstrumentHTTP..I just want something that solves all our problems regarding basic traced data on HTTP..but maybe that is bad...and in that case we would need a different package like organization.InstrumentHTTP...but not sure we do... both designs are valid and have tradeoffs and Im on the fence on this.

guilherme-santos (dismissed):
I think we have here a nice outcome 🚀

tracing/tracing.go
katcipis: I have a feeling that slog.WithContext doesn't read well.... because even when you read as a sentence, it seems like:

"logger (slog) with the given context"

BUt that is not what we are doing, we are creating a new context, from another context, with the logger stored inside (we are not creating a "slog with the given context".

I would expect this from this method:

```
log := slog.WithContext(ctx, log)
```

It returns a slog.Logger...with something from the context (with the context).

guilherme-santos (approved):

PR #6, opened 2023-06-25
feat:tracing: add ctx keys

katcipis:
Adds a way to add org_id and trace_id on contexts and retrieve them. Also add the trace id on the context on the instrumentation code, so not only a logger instance is available, but also the trace ID.

The idea is to use this as a building block for a common publisher code, that will automatically retrieve trace ID + org ID to use on events.

### DIFF
--- a/tracing/tracing.go
+++ b/tracing/tracing.go
@@ -2,6 +2,7 @@
 package tracing
 
 import (
+	"context"
 	"net/http"
 
 	"github.com/birdie-ai/golibs/slog"
@@ -32,3 +33,31 @@ func InstrumentHTTP(h http.Handler) http.Handler {
 		h.ServeHTTP(res, req.WithContext(ctx))
 	})
 }
+
+// CtxWithTraceID creates a new [context.Context] with the given trace ID associated with it.
+// Call [CtxGetTraceID] to retrieve the trace ID.
+func CtxWithTraceID(ctx context.Context, traceID string) context.Context {
+	return context.WithValue(ctx, traceIDKey, traceID)
+}
+
+// CtxGetTraceID gets the trace ID associated with this context.
+// Return the trace ID and true if there is a trace ID, empty and false otherwise.
+func CtxGetTraceID(ctx context.Context) (string, bool) {
+	val := ctx.Value(traceIDKey)
+	if val == nil {
+		return "", false
+	}
+	traceID, ok := val.(string)
+	if !ok {
+		return "", false
+	}
+	return traceID, true
+}
+
+// key is the type used to store data on contexts.
+type key int
+
+const (
+	traceIDKey key = iota
+	orgIDKey
+)
--- a/tracing/tracing.go
+++ b/tracing/tracing.go
@@ -44,7 +44,7 @@ func CtxWithTraceID(ctx context.Context, traceID string) context.Context {
 
 // CtxGetTraceID gets the trace ID associated with this context.
 // Return the trace ID and true if there is a trace ID, empty and false otherwise.
-func CtxGetTraceID(ctx context.Context) (string, bool) {
+func CtxGetTraceID(ctx context.Context) string {
 	return ctxget(ctx, traceIDKey)
 }
 
@@ -55,8 +55,7 @@ func CtxWithOrgID(ctx context.Context, orgID string) context.Context {
 }
 
 // CtxGetOrgID gets the trace ID associated with this context.
-// Return the trace ID and true if there is a trace ID, empty and false otherwise.
-func CtxGetOrgID(ctx context.Context) (string, bool) {
+func CtxGetOrgID(ctx context.Context) string {
 	return ctxget(ctx, orgIDKey)
 }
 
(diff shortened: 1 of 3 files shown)


tracing/tracing_test.go
```diff
@@ -10,16 +11,63 @@ import (
...
+	if ok {
+		t.Fatalf("unexpected trace id: %q", got)
+	}
+	got, ok = tracing.CtxGetOrgID(ctx)
+	if ok {
+		t.Fatalf("unexpected trace id: %q", got)
+	}
+
+	ctx = tracing.CtxWithTraceID(ctx, wantTraceID)
+	ctx = tracing.CtxWithOrgID(ctx, wantOrgID)
```
katcipis, line 56: Testing each isolated would make it harder to catch bugs like using the same key for both, then only one of them works at a time, wanted to be sure that using both together works fine since they are the only general purpose metadata we are going to have on our systems.

Still there is no way to extract org ID automatically since we don't have a header for that, code will have to do that manually for now since it comes on request bodies.

tracing/tracing.go
```diff
@@ -25,10 +26,56 @@ func InstrumentHTTP(h http.Handler) http.Handler {
...
+
+// CtxWithTraceID creates a new [context.Context] with the given trace ID associated with it.
+// Call [CtxGetTraceID] to retrieve the trace ID.
+func CtxWithTraceID(ctx context.Context, traceID string) context.Context {
+	return context.WithValue(ctx, traceIDKey, traceID)
+}
+
+// CtxGetTraceID gets the trace ID associated with this context.
+// Return the trace ID and true if there is a trace ID, empty and false otherwise.
+func CtxGetTraceID(ctx context.Context) (string, bool) {
```
katcipis, line 47: Instead of N functions an alternative design would be to make the keys public and the client just pass the key they want, was on the fence on this. This seemed like a more encapsulated solution (at the cost of having N functions, could get messy for a big N, but we don't plan on having a big N).

guilherme-santos: Yep, I agree. I don't expect to have a lot of keys in our context (at least in the library). Something that I'd expect though, it's to allow users to inject their own keys, but still not sure if we have something to add here or not. 

Maybe explaining a bit more, as trace_id and org_id, we may want to allow users to inject some stuff on the logs and the context. But I'd wait a bit to come up with a solution as still not clear to me what would be the use-cases. And so far seems the log would be more important, but we already have ways to add fields there, so I think we're fine.

tracing/tracing.go
```diff
@@ -25,10 +26,56 @@ func InstrumentHTTP(h http.Handler) http.Handler {
...
+
+// CtxWithOrgID creates a new [context.Context] with the given organization ID associated with it.
+// Call [CtxGetOrgID] to retrieve the organization ID.
+func CtxWithOrgID(ctx context.Context, orgID string) context.Context {
+	return context.WithValue(ctx, orgIDKey, orgID)
+}
+
+// CtxGetOrgID gets the trace ID associated with this context.
+// Return the trace ID and true if there is a trace ID, empty and false otherwise.
+func CtxGetOrgID(ctx context.Context) (string, bool) {
```
guilherme-santos, line 59: I have mixed feelings about the bool. I'd say that is unlikely to return false, and we should try to make sure that the fields are available. Regarding usage can make a little harder to use as you cannot inline it.

guilherme-santos (approved):
Just a minor comment about the bool

tracing/tracing.go
guilherme-santos: Also, zero value for string would be perfect fine to check if we need

tracing/tracing.go
katcipis: On "Something that I'd expect though, it's to allow users to inject their own keys, but still not sure if we have something to add here or not." the context API itself already allows it...so not sure what do you mean :thinking: (we define a private type exactly to avoid clashing, so users could even store the trace/org ID on the context with some other key if they want, for some reason/integration/etc).

On adding things to logs, I tried going through the path of automatically extracting things from the context, I don't think that scaled well and requires custom handlers on the logger. Seems easier to just store/retrive logger instances and manage keys on the logger.

For other purposes, then we have these two functions or the user can just add any key/value they want to contexts, this functions are just helpers that will be used together with the middleware, but don't limit what people can do AFAIK.

tracing/tracing.go
katcipis: I usually like the semantics from maps where I can always be sure if the zero value was inserted there or if it is an actual default zero value, but for maps you can always also ignore the boolean when you dont want (Go cheats =P...it can do cool stuff you cant). But in this case it does seem that inserting an empty string would be something very odd anyway (getenv has similar behavior and an alternative that returns a boolean, because sometimes it makes a difference if it was exported with empty or undefined etc).

tracing/tracing.go
guilherme-santos: I meant that we have a pattern here where we want to extract an information from the header, inject it in the context, add it to the logs, and perhaps forward it to an outgoing request.

I don't think we will need much more than traceID and orgID for the moment, but I feel that application would need to have such requirements as well. UserID could be a candidate.

So, that's what I meant by expectation, I don't have enough information but perhaps we'll need to reimplement all of this for each use-case inside of different services. But again, just thinking out loud, let's discuss when we need it and IF we need it.

tracing/tracing.go
guilherme-santos: Yep, having boolean being optional would be a nice thing. I also like the map as it's your decision if the zero value is relevant to your context or not.

guilherme-santos (approved):

PR #7, opened 2023-06-28
feat:event: add new module to help with events handling

katcipis:
For now adding the bare basics of event publishing.


event/event.go
```diff
@@ -0,0 +1,52 @@
...
+
+// Publisher represents a publisher of events of type T.
+// The publisher guarantees that the events conform to our basic schema for events.
+type Publisher[T any] struct {
+	name  string
+	topic *pubsub.Topic
+}
+
+// Body represents the general structure of the body of events.
+type Body[T any] struct {
```
guilherme-santos, line 20: Wondering if we need generics here. We'll call json.Marshal over the event, right? so interface{} / any wouldn't be enough? Can we have any guarantees using T?

katcipis: You have the guarantee that you won't send on the same publisher/topic different/arbitrary data. Once built for type T it will only send type T.

katcipis: Of course you still could create multiple publishers with different types with the same topic, but that is more explicit. Send on an any something wrong is a much more possible/honest mistake.

guilherme-santos (dismissed):
Nice! 👏🏻

event/event_test.go
```diff
@@ -0,0 +1,125 @@
...
+		orgID     = "org-id"
+	)
+
+	publisher := event.NewPublisher[Event](eventName, topic)
+	wantEvt := Event{
+		Field: fieldData,
+	}
+
+	go func() {
+		// tracing info stored no the context is propagated to the events.
```
katcipis, line 49:
```suggestion
		// tracing info stored on the context is propagated to the events.
```

guilherme-santos (approved):

PR #8, opened 2023-06-30
feat: add raw subscription

katcipis:
Add a first raw subscription that delivers any message. We will build a more specific one on top of this that will handle our events envelopes. A follow up PR will add the subscription that we will usually use more, this one is suitable for legacy events that don't follow our new envelope format.

### DIFF
--- a/event/event.go
+++ b/event/event.go
@@ -101,11 +101,13 @@ func (r *RawSubscription) Serve(handler RawMessageHandler) error {
 				<-semaphore
 			}()
 			err := handler(msg.Body)
-			if err == nil {
-				msg.Ack()
-			} else {
-				msg.Nack()
+			if err != nil {
+				if msg.Nackable() {
+					msg.Nack()
+				}
+				return
 			}
+			msg.Ack()
 		}()
 	}
 }
--- a/event/event.go
+++ b/event/event.go
@@ -25,6 +25,21 @@ type Body[T any] struct {
 	Event   T      `json:"event"`
 }
 
+// Message represents a raw message received on a subscription.
+type Message struct {
+	body []byte
+}
+
+// RawSubscription represents a subscription that delivers messages as is.
+// No assumptions are made about the message contents. This should rarely be used in favor of [Subscription].
+type RawSubscription struct {
+	sub            *pubsub.Subscription
+	maxConcurrency int
+}
+
+// RawMessageHandler is responsible for handling raw messages from a subscription.
+type RawMessageHandler func(Message) error
+
 // NewPublisher creates a new event publisher for the given event name and topic.
 func NewPublisher[T any](name string, t *pubsub.Topic) *Publisher[T] {
 	return &Publisher[T]{
@@ -51,16 +66,6 @@ func (p *Publisher[T]) Publish(ctx context.Context, event T) error {
 		Body: encBody,
 	})
 }
-
-// RawSubscription represents a subscription that delivers messages as is.
-// No assumptions are made about the message contents. This should rarely be used in favor of [Subscription].
-type RawSubscription struct {
-	sub            *pubsub.Subscription
-	maxConcurrency int
-}
-
-// RawMessageHandler is responsible for handling raw messages from a subscription.
-type RawMessageHandler func([]byte) error
 
 // NewRawSubscription creates a new raw subscription. It provides messages in a
 // service like manner (serve) and manages concurrent execution, each message
@@ -100,7 +105,7 @@ func (r *RawSubscription) Serve(handler RawMessageHandler) error {
 			defer func() {
 				<-semaphore
 			}()
-			err := handler(msg.Body)
+			err := handler(Message{body: msg.Body})
 			if err != nil {
 				if msg.Nackable() {
 					msg.Nack()
@@ -116,4 +121,9 @@ func (r *RawSubscription) Serve(handler RawMessageHandler) error {
 // The subscription should not be used after this method is called.
 func (r *RawSubscription) Shutdown(ctx context.Context) error {
 	return r.sub.Shutdown(ctx)
+}
+
+// Body of the message.
+func (m Message) Body() []byte {
+	return m.body
 }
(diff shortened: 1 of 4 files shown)


event/event.go
```diff
@@ -50,3 +51,62 @@ func (p *Publisher[T]) Publish(ctx context.Context, event T) error {
...
+// Serve will start serving all messages from the subscription calling handler for each
+// message. It will run until [RawSubscription.Shutdown] is called.
+// If the error is nil Ack is sent.
+// If a non-nil error is returned by the handler Unack will be sent.
+// Serve may be called multiple times, each time will start a new serving service that will
+// run up to "maxConcurrency" goroutines.
+func (r *RawSubscription) Serve(handler RawMessageHandler) error {
+	semaphore := make(chan struct{}, r.maxConcurrency)
+	for {
+		semaphore <- struct{}{}
```
katcipis, line 92: Interesting enough the semaphore example from the lib seems broken: https://pkg.go.dev/gocloud.dev@v0.30.0/pubsub#example-Subscription.Receive-Concurrent

It first receives a message and then acquires the semaphore. If you have to wait for minutes on the semaphore the message will timeout since you acquired the message first. Makes more sense to first of all acquire the semaphore, then do anything.

Unless Im missing something.

guilherme-santos: Yep, it seems so. they are doing the semaphore to handle the message not to read the message.

event/event.go
```diff
@@ -50,3 +51,69 @@ func (p *Publisher[T]) Publish(ctx context.Context, event T) error {
...
+			}
+			msg.Ack()
+		}()
+	}
+}
+
+// Shutdown will shutdown the subscriber, stopping any calls to [RawSubscription.Serve].
+// The subscription should not be used after this method is called.
+func (r *RawSubscription) Shutdown(ctx context.Context) error {
+	return r.sub.Shutdown(ctx)
```
guilherme-santos, line 118: Our use-cases is slighly different then the example, but don't we need to wait for all messages that are being handling? Or shutdown would solve that?

event/event.go
```diff
@@ -50,3 +51,69 @@ func (p *Publisher[T]) Publish(ctx context.Context, event T) error {
...
+
+// RawSubscription represents a subscription that delivers messages as is.
+// No assumptions are made about the message contents. This should rarely be used in favor of [Subscription].
+type RawSubscription struct {
+	sub            *pubsub.Subscription
+	maxConcurrency int
+}
+
+// RawMessageHandler is responsible for handling raw messages from a subscription.
+type RawMessageHandler func([]byte) error
```
guilherme-santos, line 63: Just worred if is this not too less? Don't we need to inject the log even for this case? I'm also afraid that this wouldn't allow us to add more things that may appear in the future, do you think using a struct perhaps would give more ways to extend?

In theory, our regular subscription will use the RawMessaHandler, right? Can we build everything on top of the []byte?

event/event.go
katcipis: Shutdown will flush all pending acks: https://pkg.go.dev/gocloud.dev@v0.30.0/pubsub#Subscription.Shutdown

But it won't wait for all outstanding messages to be handled. We could implement this behavior here but Im not sure we need it (messages are just sent back to the subscription and retried on a different process, unless this happens all the time it should not cause major issues).

event/event.go
katcipis: Right now we can. I thought about this and was on the fence. We can't create a logger right now because there is no context that is useful to add on it, it would have the same effect as getting a default one. But I agree that it would make it more future proof..but at the same time this library is v0 and changing it should be fairly simple and then code using it would just need to do a small fix, so I started with the extreme minimum (but maybe I was too extreme ? :laughing: )

I'm OK with creating our own Message type and it would only have a Body for now. Maybe I was too Conan here.



WDYT @lfritz ?

event/event.go
guilherme-santos: Mmh, this wouldn't be a problem in an environment where we have idempotency, which, AFAIK it's not our case. we can call multiple times the same API, publish the same event multiple time,s and so on.

This can happen anyway, but if we don't handle it feels like we're increasing the probabilities as any new deployment we're adding the change of happening increase.

We can treat it as an improvement, though, I just feel should be something that we should have before we integrate it everywhere.

katcipis: which events we have that if processed twice causes issues ? :thinking: AFAIK it would only generate unnecessary reprocessing in some cases..but no severe issues (like debiting money twice etc). When reprocessing is an issue I prefer to have explicit idempotency mechanisms, trusting the broker to deliver only once is tricky IMHO.

For example, google has it: https://cloud.google.com/pubsub/docs/exactly-once-delivery
And yet depending on what happens on the handlers, it may be delivered to more than one (after the ack timeout), so if you are doing something that could cause problems if done twice you will need to handle it on the code anyway..or have severe problems.

+ If this is as essential as you say it doesn't change the interface of shutdown..so adding it later and updating the libraries should be fairly OK too. So not so sure we need this right away :thinking:.

guilherme-santos: Yep, I'm aware it can happen anyway. I just feel that not handling it increases the probability of happening on every deployment. But I also acknowledge that it can be treated as improvement, so all good!

katcipis: To be on the safe side we can add this right away, can do in a separate PR then.

event/event.go
katcipis: Added a event.Message type for now, WDYT @guilherme-santos ?

event/event.go
```diff
@@ -24,6 +25,21 @@ type Body[T any] struct {
 	Event   T      `json:"event"`
 }
 
+// Message represents a raw message received on a subscription.
+type Message struct {
+	body []byte
```
guilherme-santos, line 30: This can also be used by the outside, should be public, shouldn't it?

katcipis: It is accessed through a method to guarantee it is read only/immutable.

guilherme-santos (approved):

event/event.go
katcipis: I usually declare all types and then start declaring the methods..but it is only an old habit.

PR #9, opened 2023-07-02
feat: add subscription

katcipis:
Adding event.Subscription that will handle envelope data for events. It constrains for all events to have the same data/schema and validates the name of the event matches the one expected. Invalid data is discarded automatically. To handle more flexible scenarios use RawSubscription.

Since we don't want to focus on allowing different events on the same subscription there is no support for it for now. If it becomes necessary my idea would be to build a different Subscription that supports muxing like the http.ServeMux but multiplexing event names instead of URL paths.

Depends on #8 (so still draft).

### DIFF
--- a/event/event.go
+++ b/event/event.go
@@ -37,6 +37,11 @@ type Subscription[T any] struct {
 // by using [slog.FromCtx].
 type EventHandler[T any] func(context.Context, T) error
 
+// Message represents a raw message received on a subscription.
+type Message struct {
+	body []byte
+}
+
 // RawSubscription represents a subscription that delivers messages as is.
 // No assumptions are made about the message contents. This should rarely be used in favor of [Subscription].
 type RawSubscription struct {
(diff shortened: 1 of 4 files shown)


event/event.go
```diff
@@ -102,6 +164,7 @@ func (r *RawSubscription) Serve(handler RawMessageHandler) error {
 			}()
 			err := handler(msg.Body)
 			if err != nil {
+				slog.Error("message handler failed", "message", string(msg.Body), "error", err)
```
katcipis, line 167: I dont like logging in libraries, but in some cases the error will be produced internally by the library, the user code won't even be called, so there will be no opportunities to log anything.

event/event.go
```diff
@@ -80,6 +104,44 @@ func NewRawSubscription(url string, maxConcurrency int) (*RawSubscription, error
...
+	return s.rawsub.Serve(func(msg []byte) error {
+		var event Envelope[T]
+
+		if err := json.Unmarshal(msg, &event); err != nil {
+			return fmt.Errorf("parsing event as JSON, event: %v, error: %v", string(msg), err)
+		}
+
+		if event.Name != s.name {
+			return fmt.Errorf("event name doesn't match %q: event: %v", s.name, string(msg))
+		}
```
guilherme-santos, line 125: This is a publisher error, right? For such case, we're nacking and sending to the DLQ?

guilherme-santos (approved):

PR #10, opened 2023-07-03
chore: add metrics spec

katcipis:
Specify metrics we want to export on the event library. PR with code will follow.


event/README.md
```diff
@@ -3,3 +3,52 @@
...
+Total of published messages.
+
+Labels:
+
+* status : "ack" or "nack".
+* name : name of the event.
+
+### Subscription
+
+#### event_process_duration_seconds : histogram
```
katcipis, line 38: Not a huge fan of this metric name x_x.

guilherme-santos (dismissed):
🚀

event/README.md
guilherme-santos: What bothers you? we could use publish/consume WDYT?

matmarzolagomes (approved):

event/README.md
katcipis: Not sure what was bothering me..but was feeling off with the name :laughing: . I thought about consume but it also didn't made me happy...maybe nothing will made me happy here :laughing: 

WDYT @lfritz ?

lfritz: This is basically the time it takes until we can ack or nack the message, right? Consume and process both seem fine for that case, in my mind :smile:

event/README.md
```diff
@@ -3,3 +3,52 @@
...
+### Publisher
+
+#### event_publish_duration_seconds : histogram
+
+Measure publish duration time for each event.
+Useful when debugging connectivy issues between a service and the message broker.
+
+Labels:
+
+* status : "ack" or "nack".
```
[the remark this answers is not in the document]
katcipis: yeah I tried to be more event/pubsub specific but it doesn't fit well with status (unless we rename the label).

event/README.md
katcipis: will leave as process for now

guilherme-santos (approved):
👍🏻

PR #11, opened 2023-07-04
fix: update slog and fix handler creation

guilherme-santos:


guilherme-santos:
Talked to @katcipis and he prefer to wait to migrate when the package is moved to stdlib

katcipis (approved):

PR #12, opened 2023-07-05
Expose ParseLevel and ParseFormat

guilherme-santos:


perebaj (approved):

slog/slog.go
```diff
@@ -184,7 +184,7 @@ const (
 	loggerKey key = iota
 )
 
-func validateLogLevel(level string) (Level, error) {
+func ParseLevel(level string) (Level, error) {
```
katcipis, line 187: Maybe add some documentation for them since now they are public. Just curious why are they being made public ? The LoadConfig function is not working on some scenario ?

guilherme-santos: Sorry didn't think about it. Added the doc and add also a testcase.

Yeah, I'm reading other envvars on my code already, so I thought I should read everything and just send the config. Just a different approach I guess.

slog/slog_test.go
```diff
@@ -76,6 +76,67 @@ func TestDefaultLoggerFromContext(t *testing.T) {
 	}
 }
 
+func TestParseLevel(t *testing.T) {
```
perebaj, line 79: ✅

perebaj (approved):

guilherme-santos:
@katcipis I'm going to merge it as it's blocking me, we can revisit later the idea of creating a envvar module.

PR #13, opened 2023-07-06
feat:event: add metric instrumentation

katcipis:


event/metrics.go
```diff
@@ -0,0 +1,91 @@
...
+// MustRegisterMetrics will register all event related metrics on the given registry.
+// If metrics with the same name already exist no the register this function will panic.
+func MustRegisterMetrics(registry *prometheus.Registry) {
+	registry.MustRegister(publishDuration, publishCounter, processCounter, processDuration)
+}
+
+// SampledMessageHandler will instrument the given MessageHandler returning a new one
+// that samples metrics. These will be `event_process_*` metrics using as `name` the
+// given eventName.
+func SampledMessageHandler(handler MessageHandler, eventName string) MessageHandler {
```
katcipis, line 18: Not sure this is the best name, open to alternatives. The idea is that now anyone using the RawSubscription can also instrument metrics easily + provide an eventName and the same API is used by the Subscription (which has a event name).

guilherme-santos (approved):
Top!

event/metrics.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	labels := prometheus.Labels{
+		"status": status,
+		"name":   name,
+	}
+	processDuration.With(labels).Observe(elapsed.Seconds())
+	processCounter.With(labels).Inc()
+}
+
+var (
```
perebaj, line 54: Doubt: Is it normal to define var in the end of the file?

event/metrics.go
```diff
@@ -0,0 +1,91 @@
...
+	publishDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name: "event_publish_duration_seconds",
+			Help: "Duration of event publish",
+			Buckets: []float64{
+				.1, .2, .3, .4, .5, .6, .7, .8, .9, 1,
```
perebaj, line 60: Just an interest post about how histogram are calculated and why we need to set up buckets between the aimed values:
https://twitter.com/BdKozlovski/status/1675557351493890048?s=20
I never thought about it, maybe can be nice !

event/metrics.go
perebaj: for me the name sounds good

perebaj (approved):
For me, looks good. But I'm not a Golang 🧙

event/metrics.go
katcipis: Not normal, there is no default/standard way to do this in Go. I like starting the file with everything that is public and leaving everything that is private after the public things, but that is not something that everyone does..it is something that I like. I'm a very top/down person and when I read code I like to see public interfaces/types/functions first and then to start digging in to internals. But again, not a normal thing...just my thing =P.

guilherme-santos:
@perebaj 

Wrong sentence:
> For me, looks good. But I'm not a Golang 🧙

Right sentence

> For me, looks good, but I'm still becoming a **Gopher** 🧙

😅

event/metrics.go
guilherme-santos: I also like this approach actually... Most important things on top and going down with more specific/less important... Also I like to define methods related to a struct close to where the struct is defined

PR #14, opened 2023-07-16
chore: add golang CI config

katcipis:
Detect undocumented public symbols and some other issues like stuttering.

### DIFF
--- a/event/event.go
+++ b/event/event.go
@@ -35,7 +35,7 @@ type (
 		rawsub *MsgSubscription
 	}
 
-	// SubscriptionHandler is responsible for handling events from a subscription.
+	// SubscriptionHandler is responsible for handling events from a [Subscription].
 	// The context passed to the handler will have all metadata relevant to that
 	// event like org and trace IDs. It will also contain a logger that can be retrieved
 	// by using [slog.FromCtx].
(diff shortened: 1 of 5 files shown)


event/event.go
```diff
@@ -28,16 +28,18 @@ type (
...
 
+	// Subscription is a subscription that received only specific types of events
+	// defined by T.
 	Subscription[T any] struct {
 		name   string
 		rawsub *MsgSubscription
 	}
 
-	// EventHandler is responsible for handling events from a subscription.
+	// SubscriptionHandler is responsible for handling events from a [Subscription].
```
lfritz, line 38: You didn't want to just call it `Handler` (making it `event.Handler`)?

lfritz (dismissed):

event/event.go
katcipis: I thought about it, it was my first option, but we also have the MsgSubscription and its Handler...so I was on the fence if it was not going to be too vague.

katcipis: Like on event we have Publisher/Subscription/MsgSubscription...it felt like Handler would not be specific enough...but it does read very well =P

katcipis: WDYT @guilherme-santos ?

guilherme-santos: Yeah, I like how it reads as well. Maybe MsgSubscription is not great.

As the struct is called `Message`, `Handler` and `MessageHandler` wouldn't be good?

katcipis: hmm yeah, so it would be MessageHandler => Message and Handler -> event.Envelope[T] ?

katcipis: hmmm I really like just event.Handler and it would work perfectly without the Message stuff.... will go with this name and hope we can remove the old stuff someday :laughing:

guilherme-santos (approved):
🙏🏻

PR #15, opened 2023-07-16
feat: add service module

katcipis:
For now it only has some very basic logic for shutting down multiple services gracefully that can be quite useful. Services using this will block their main thread by calling handler.Wait(). Then if any cancellation/signal is received everything gets gracefully shutdown :-).

### DIFF
--- a/service/shutdown.go
+++ b/service/shutdown.go
@@ -3,10 +3,10 @@ package service
 
 import (
 	"context"
-	"sync"
+	"errors"
 	"time"
 
-	"github.com/birdie-ai/golibs/slog"
+	"github.com/sourcegraph/conc/pool"
 )
 
 // Shutdowner represents a service that can shutdown.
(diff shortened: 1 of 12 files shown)


service/shutdown.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/sourcegraph/conc/pool"
+)
+
+// Shutdowner represents a service that can shutdown.
+type Shutdowner interface {
+	Shutdown(context.Context) error
+}
+
+// ShutdownHandler handles the shutdown of multiple services.
+// It waits for a context to be cancelled to then call all added services Shutdown methods.
```
lfritz, line 18:
```suggestion
// It waits for a context to be cancelled to then call each service's Shutdown method.
```

service/shutdown.go
```diff
@@ -0,0 +1,56 @@
...
+// Must be called before [ShutdownHandler.Wait] is called.
+func (s *ShutdownHandler) Add(service Shutdowner) {
+	s.services = append(s.services, service)
+}
+
+// Wait will wait for the given [ctx] to be cancelled.
+// When [ctx] is cancelled it will shutdown all services
+// concurrently and wait for all of them to finish before returning.
+// It will wait for each service to shutdown for the wait period provided on
+// NewShutdownHandler.
```
lfritz, line 39:
```suggestion
// Wait will wait for the given [ctx] to be cancelled.
// When [ctx] is cancelled it will shut down all services
// concurrently and wait for all of them to finish before returning.
// It will wait for each service to shut down for the wait period provided on
// NewShutdownHandler.
```

lfritz (approved):

guilherme-santos (dismissed):

service/shutdown.go
```diff
@@ -0,0 +1,56 @@
+// Package service provides functionality common to services
+package service
+
+import (
+	"context"
+	"errors"
+	"time"
+
+	"github.com/sourcegraph/conc/pool"
```
perebaj, line 9: When we are dealing with Golang concurrency is safe to use third libraries that abstract some concepts like this one?

perebaj (approved):

service/shutdown.go
katcipis: Depends on the library and what you are doing :-). I started using conc for a while and so far didn't suffered from any bugs (code looks pretty good overall too). But there is always a chance the library will screw you..so if it is something very simple prefer the stdlib, but for something kinda like a concurrent map/reduce this library makes it fairly easier.

katcipis: Also..for the sake of learning..when playing around on personal project I would advocate using only the stdlib/waitgroups/channels/mutexes so you can learn the basic building blocks. But more high level ones can be useful, golang/x has something similar with errgroup: https://pkg.go.dev/golang.org/x/sync/errgroup, but conc is more expressive since it leverages parametric typing.

Some stuff on conc I find quite complex/cumbersome..not sure I would use everything..but I do like the Pools.

PR #17, opened 2023-07-21
chore: add Fatal and FatalCtx

guilherme-santos:


slog/slog.go
```diff
@@ -155,6 +161,12 @@ func ErrorCtx(ctx context.Context, msg string, args ...any) {
 	slog.ErrorCtx(ctx, msg, args...)
 }
 
+// FataCtx is equivalent to ErrorCtx() followed by a call to os.Exit(1).
+func FataCtx(ctx context.Context, msg string, args ...any) {
```
katcipis, line 165:
```suggestion
// FatalCtx is equivalent to ErrorCtx() followed by a call to os.Exit(1).
func FatalCtx(ctx context.Context, msg string, args ...any) {
```

katcipis (changes requested):
Good idea, typo on the function name :heart:

slog/slog.go
guilherme-santos: Opsie.

katcipis (approved):

lfritz (approved):

PR #19, opened 2023-07-24
:sparkles: style(metrics): README

perebaj:


event/README.md
```diff
@@ -6,18 +6,21 @@ to the events.
...
+The library also provides metrics both for publishers and subscribers. We use Prometheus for metrics.
+
 The metrics are always sampled but are not registered by default anywhere, so if you want the metrics
-you need to opt-in on them by calling [event.RegisterMetrics](TODO_LINK).
+you need to opt-in on them by calling [event.MustRegisterMetrics](TODO_LINK).
+
+It's possible to find some examples of how to configure it in your Golang service based on these repositories, [feedback-extractor](https://github.com/birdie-ai/feedback-extraction-service/blob/main/cmd/feedback-extraction-service/main.go) & [aggregator](https://github.com/birdie-ai/aggregator/blob/main/cmd/aggregator/main.go)
 
-Here we document the metrics generated.
+Don't forget to expose the metrics port, more details about it, [Metrics](https://github.com/birdie-ai/feedback-extraction-service/blob/main/cmd/feedback-extraction-service/main.go)
```
katcipis, line 16: I would not add this here, the library is open and I don't think the library is responsible for reminding you to export metrics :thinking:.

If you really wan't to...then we could add a code snippet on how to register the metrics + running a server.

event/README.md
```diff
@@ -6,18 +6,21 @@ to the events.
...
 
 ### Publisher
 
 #### event_publish_duration_seconds : histogram
 
 Measure publish duration time for each event.
-Useful when debugging connectivy issues between a service and the message broker.
+Useful when debugging connectivity issues between a service and the message broker.
```
katcipis, line 23: nice :100:

event/README.md
```diff
@@ -6,18 +6,21 @@ to the events.
 
 ## Metrics
 
-The library also provides metrics both for publishers and subscribers. We use prometheus for metrics.
+The library also provides metrics both for publishers and subscribers. We use Prometheus for metrics.
+
 The metrics are always sampled but are not registered by default anywhere, so if you want the metrics
-you need to opt-in on them by calling [event.RegisterMetrics](TODO_LINK).
+you need to opt-in on them by calling [event.MustRegisterMetrics](TODO_LINK).
```
katcipis, line 12: nice !

event/README.md
lfritz: Maybe it could go in the Handbook instead? https://github.com/birdie-ai/handbook/blob/main/monitoring.md#metrics
There is some Python-specific documentation there, we could have something similar for Go.

event/README.md
```diff
@@ -6,18 +6,21 @@ to the events.
...
+
+It's possible to find some examples of how to configure it in your Golang service based on these repositories, [feedback-extractor](https://github.com/birdie-ai/feedback-extraction-service/blob/main/cmd/feedback-extraction-service/main.go) & [aggregator](https://github.com/birdie-ai/aggregator/blob/main/cmd/aggregator/main.go)
```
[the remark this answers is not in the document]
perebaj: What you mean with snippet?

something like
```Go
func main() {
//implementation here?
}


```

event/README.md
guilherme-santos: I'm not sure if we should put in the handbook. Handbook should be general stuff, doc on how to use the library makes more sense, here, no?

event/README.md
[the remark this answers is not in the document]
guilherme-santos: Yep! It's also pointing to main branch, if someone change something the links will break. +1 for snippets

event/README.md
perebaj: When Katz and I were trying to improve the python repositories, we tried to make python driven documentation, for example

[logging](https://github.com/birdie-ai/handbook/blob/main/logging/python.md)
[metrics](https://github.com/birdie-ai/handbook/blob/main/metrics/python.md)

but we do not give up general documentation [General Monitoring](https://github.com/birdie-ai/handbook/blob/main/monitoring.md)

so, I think that the handbook it's a good place to expose the ideas around of how to do something, and some code examples, in this specific case, the example is using a library maintaining by us. 
for this reason +1 for keep on handbook

event/README.md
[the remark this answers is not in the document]
perebaj: ok :mage

event/README.md
katcipis: +1 to moving to the handbook. We could have examples here on how to use the library, but I would not add specific documentation on how to deploy things or opinionated ways on how to use it, that could be on the handbook indeed (we already have a lot of how-to's there).

For example, giving examples on how to expose prometheus metrics, this is not specific to the library actually...anyone using the library can expose metrics whatever the way they want. The fact that we run a separate server in a different port etc is something that should be documented on the handbook..not here (including internal examples on how to do it).

event/README.md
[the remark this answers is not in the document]
katcipis: Yeah snippet would be code snippet :-)

event/README.md
guilherme-santos: Mmh, now I go to [the library I want to use](https://github.com/birdie-ai/golibs/tree/improve_metrics_readme/event) I have no example of how to use it, what I need to know, and so on.

I'm not saying to we shouldn't have anything on handbook, but I seel more general stuff, that is not related to a specific library, e.g. how to name metrics, which fields should have in the log, and so on. Because we can have different implementations for different languages.

Also, something that I feel weird  is the fact that I have to know that examples to use this library is somewhere else, as I agree doesn't make sense to link here to the handbook but then I think is a must we have more docs here on how to use _this_ library.

katcipis: I would be OK with examples here, what is odd is talk about how to run a http.Server providing metrics (not specific to the library, this is a detail on how to integrate prometheus, depends on your setup). For example, it is our current standard to always export metrics in a separate server in a separate port, isolated from the service, our examples on the handbook will be like that because this is how we do it. But someone could just add a /metrics on their current API and it would also work..if they feel like it.

Im not against having a few examples on the library, I just don't think the library is how we teach how birdie code should look like, for anything that is more about standards or how we run things (like mentioning kubernetes, examples of YAML etc) I would put on the handbook.

Maybe we are thinking different things.

katcipis: The final solution in my opinion will be an http library that does these kind of things and make it even simpler (metrics server and also metrics instrumentation) + the handbook will document that we have those libraries, were to find them, etc and maybe include some examples (with links to private projects), because it is our main entry point for engineers, when you don't know how to do something you should look there first, it is the hub (at least it is intended to be).

PR #20, opened 2023-07-27
:sparkles: feat: infof

perebaj:


katcipis (commented):
I would argue against formatting, there is a reason slog doesnt have it, when you are doing structured logging there is no reason to do formatting, you use the field structure. For very rare exceptions I would just do manually, but not on the lib.

There are some good ideas here, including rationale around formatting:

https://engineering.hellofresh.com/logging-rules-of-thumb-f6c0f71a2351

PR #22, opened 2023-08-08
feat: add service build info metrics

katcipis:


service/metrics.go
```diff
@@ -0,0 +1,45 @@
...
+func SampleBuildInfo() {
+	goVersion := "undefined"
+	revision := "undefined"
+
+	goBuildInfo, ok := debug.ReadBuildInfo()
+	if ok {
+		goVersion = goBuildInfo.GoVersion
+		for _, buildSetting := range goBuildInfo.Settings {
+			if buildSetting.Key == "vcs.revision" {
+				revision = buildSetting.Value
```
guilherme-santos, line 25: This is the go revision, right? Shouldn't we also add our service version?

katcipis: not it is not, this is our service version (from vcs/git). The go version is goBuildInfo.GoVersion.

katcipis: Reference:https://pkg.go.dev/runtime/debug#BuildSetting

"vcs.revision: the revision identifier for the current commit or checkout"

guilherme-santos: Mmh, shit! I always injected in the makefile, I didn't know about that.

guilherme-santos (approved):

service/metrics.go
katcipis: All good :-), this is new stuff that came on Go 1.18 (which now it is not that new actually...getting old sux :rofl: )

PR #23, opened 2023-08-09
fix: update golibs

guilherme-santos:
I think this is one downside of having multiple modules.


lfritz (approved):

katcipis:
> I think this is one downside of having multiple modules.

lets put them all together in the same module then

PR #25, opened 2023-08-10
refactor: make golibs a single module

katcipis:


lfritz (approved):
Are you going to do a release of the new one-module golibs after merging?

katcipis:
> Are you going to do a release of the new one-module golibs after merging?

Yeah that is the idea, the separation into multiple independent modules was premature IMHO.

katcipis:
But older tags/module references will keep working, so should not break anything.

guilherme-santos (approved):

PR #26, opened 2023-08-15
chore: add log when event is malformed

guilherme-santos:


event/event.go
```diff
@@ -134,18 +134,22 @@ func (s *Subscription[T]) Serve(handler Handler[T]) error {
...
+		ctx := context.Background()
+		log := slog.FromCtx(ctx)
+
 		if err := json.Unmarshal(msg.Body(), &event); err != nil {
+			log.Error("unable to parse event as JSON", "error", err, "event", msg)
 			return fmt.Errorf("parsing event as JSON, event: %v, error: %v", msg, err)
 		}
 
 		if event.Name != s.name {
+			log.Error("event name doesn't match handler", "event_expected", s.name, "event_received", event.Name)
```
katcipis, line 146:
```suggestion
			log.Error("event name doesn't match handler", "expected", s.name, "received", event.Name)
```

Given that the message is already clear about it being the even name.

event/event.go
```diff
@@ -134,18 +134,22 @@ func (s *Subscription[T]) Serve(handler Handler[T]) error {
...
 			return fmt.Errorf("event name doesn't match %q: event: %v", s.name, msg)
 		}
 
-		ctx := tracing.CtxWithTraceID(context.Background(), event.TraceID)
+		ctx = tracing.CtxWithTraceID(ctx, event.TraceID)
```
katcipis, line 150: Nice !

katcipis (dismissed):

event/event.go
guilherme-santos: Yeah, I wasn't sure about it

PR #28, opened 2023-08-18
feat: add basic http instrumentation/metrics

katcipis:
To avoid duplicating the basic metric instrumentation code + help us export consistent metrics across different services, at least for basic HTTP stuff (services still can have optional domain specific metrics).

### DIFF
--- a/service/metrics.go
+++ b/service/metrics.go
@@ -1,14 +1,40 @@
 package service
 
 import (
+	"net/http"
 	"runtime/debug"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
 // MustRegisterMetrics will register all metrics on the given registry.
 func MustRegisterMetrics(registry *prometheus.Registry) {
 	registry.MustRegister(buildInfo)
+}
+
+// MustRegisterHTTPMetrics will register all HTTP related metrics on the given registry.
+func MustRegisterHTTPMetrics(registry *prometheus.Registry) {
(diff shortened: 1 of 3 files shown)


service/metrics.go
```diff
@@ -1,16 +1,42 @@
...
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
 // MustRegisterMetrics will register all metrics on the given registry.
 func MustRegisterMetrics(registry *prometheus.Registry) {
 	registry.MustRegister(buildInfo)
 }
 
+// MustRegisterHTTPMetrics will register all HTTP related metrics on the given registry.
+func MustRegisterHTTPMetrics(registry *prometheus.Registry) {
```
katcipis, line 17: Separated since a few services may not expose HTTP or expose other protocols in the future.

service/metrics_test.go
```diff
@@ -1,15 +1,36 @@
...
+	handler = service.InstrumentHTTPByPath(handler, "/test")
+
+	req := httptest.NewRequest(http.MethodGet, "/test", nil)
+	res := httptest.NewRecorder()
+
+	handler.ServeHTTP(res, req)
+
+	if !called {
+		t.Fatal("want instrumented handler to be called")
+	}
```
katcipis, line 35: Not an extremely complete test, it is also possible to test the metrics themselves using the prometheus library, but I didn't find a quick/nice example involving the collector interface, so I focus on at least showing that basic behavior is not broken because of the metrics.

guilherme-santos (dismissed):
Nice, thanks!

katcipis:
Was missing the docs.

katcipis:
> > Was missing the docs.
> 
> Shame on you!

PR #29, opened 2023-08-29
slog: add MustParseAndConfigure function

lfritz:
Currently we're repeating this code in most services.

### DIFF
--- a/slog/slog.go
+++ b/slog/slog.go
@@ -52,6 +52,35 @@ type Config struct {
 	Format Format
 }
 
+// MustParseAndConfigure calls [ParseAndConfigure] and [Fatal] in case of an error.
+func MustParseAndConfigure(level, format string) {
+	err := ParseAndConfigure(level, format)
+	if err != nil {
+		Fatal("Unable to configure slog", "error", err)
+	}
+}
+
+// ParseAndConfigure parses the log level and format and changes the default logger configuration
+// accordingly. If level or format can't be parsed, it logs a warning and uses a default value.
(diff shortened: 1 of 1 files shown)


slog/slog.go
```diff
@@ -52,6 +52,35 @@ type Config struct {
...
+		Fatal("Unable to configure slog", "error", err)
+	}
+}
+
+// ParseAndConfigure parses the log level and format and changes the default logger configuration
+// accordingly. If level or format can't be parsed, it logs a warning and uses a default value.
+func ParseAndConfigure(level, format string) error {
+	logLevel, err := ParseLevel(level)
+	if err != nil {
+		Warn("Unable to parse log level", "error", err)
```
katcipis, line 68: hmm I don't think this is desirable...won't it be confusing when you misconfigure a service and then the log level is just some other one ? (but at least you get a warning). Not sure about putting this on golibs... I would not personally implement this on a service Im maintaining..if it is misconfigured I prefer to abort, then the crash will be detected (old deployment keeps running, no downtime) and I fix the service deployment...I don't want to have a wrong configuration and then I need to catch a single warning on the logs that say it is actually badly configured.

It would be OK to provide defaults when there is no config (empty), but working with misconfigured values is odd IMHO.

lfritz: That's true, there's no good reason to provide a default here :thinking:

guilherme-santos: I feel the same actually, otherwise it can mask everything is fine

PR #32, opened 2023-09-08
Add gcloud format for httpRequest

guilherme-santos:
I want to have a log compatible with httpRequest on Google Logging


tracing/tracing.go
```diff
@@ -37,14 +37,27 @@ func InstrumentHTTP(h http.Handler) http.Handler {
...
 		start := time.Now()
 		defer func() {
 			elapsed := time.Since(start)
-			log.Info("handled request", "url", req.URL, "method", req.Method, "status", resWriter.status, "elapsed", elapsed)
+			httpRequest = append(httpRequest,
+				"status", resWriter.status,
+				"response_size", resWriter.contentLength,
+				"elapsed", elapsed.String(),
+			)
+			log.Info("handled request", slog.Group("httpRequest", httpRequest...))
```
guilherme-santos, line 60: Unfortunately, I couldn't find a way to rename the group `httpRequest`. I want to have it here `http_request` and replace only for gcloud

katcipis: This is odd, you can't intercept the "http_request" and then rename it to "httpRequest" ? (this is usually possible for any field) I never used Group...maybe define a struct with all the fields you want to log and then calling it here like:

```go
httpReq := httpRequest{
   Status: etc,
   Lala: etc,
}
log.Info("handle request", "http_request", httpReq)
```

Just wondering what could be changing the behavior and not allowing you to change the key.

guilherme-santos: Nope: the signature is for `ReplaceAttr` is `func(groups []string, a slog.Attr) slog.Attr`, I receive the list of groups, but I cannot return it... 

Also from https://pkg.go.dev/log/slog@go1.21.1#HandlerOptions

> 	// The first argument is a list of currently open groups that contain the
	// Attr. It must not be retained or modified. ReplaceAttr is never called
	// for Group attributes, only their contents.

Yep, I can check if this would work, because then is not a group but a single field!

tracing/tracing.go
```diff
@@ -94,6 +108,11 @@ func (r *responseWriter) WriteHeader(statusCode int) {
 	r.ResponseWriter.WriteHeader(statusCode)
 }
 
+func (r *responseWriter) Write(b []byte) (int, error) {
+	r.contentLength += int64(len(b))
```
katcipis, line 112: Since now we are messing with writes we could add a response body on the test (from what I can remember it is a simple status code that is tested).

tracing/tracing.go
katcipis: If you want to keep things dynamic, I think this also should work (but never used it):

```go
httpReq := map[string]any{
    "a": "b"
}
log.Info("handle request", "http_request", httpReq)
```

guilherme-santos: Not sure, I would need to test. But it seems to me, that it'll be `{"http_request": "{\"a\":\"b\"}"`, like a json inside of the string, not as different fields.

guilherme-santos: I had a use-case on search that I wanted to know what was the log level set. Maybe an approach is being able to get those information. Which is the log level/format, then I could dynamically set the right field. Just an idea

katcipis: It does work with the map when the output is JSON: https://go.dev/play/p/bKyMG2jIFc3
It gets uglier with the text handler (prints a map on the output, instead of the nice grouping).
It is just a little odd how grouping seems redundant with having a struct/map..at least for simple scenarios like these (there is the whole WithGroup that then allows some more complex idioms).

slog/slog.go
```diff
@@ -95,11 +98,32 @@ func Configure(cfg Config) error {
...
+				if a.Key == slog.MessageKey {
+					a.Key = "message"
+				}
 			}
-			if a.Key == slog.MessageKey {
-				a.Key = "message"
+
+			// Customize the http request fields
+			// More: https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#HttpRequest
+			if len(groups) > 0 && groups[0] == "httpRequest" {
```
katcipis, line 112: We may have multiple groups no ? you need to iterate on all groups to find "httpRequest" no ? (maybe that is why you didn't find the http_request ?). Even if on isolated tests we only have one group or this one is the first..Im not sure this will always be true depending on how users of the library use grouping/manipulate the loggers.

tracing/tracing.go
katcipis: Playing around further with this: https://go.dev/play/p/tSORDXDZJ1Q
seems like a rewriting using the Attribute key is possible, instead of grouping (but maybe there is some benefit to grouping that Im missing..Im not very used to this concept, I think slog is the first time I see this Group idea.. usually you only have key/values).

katcipis: > Nope: the signature is for `ReplaceAttr` is `func(groups []string, a slog.Attr) slog.Attr`, I receive the list of groups, but I cannot return it...
> 
> Also from https://pkg.go.dev/log/slog@go1.21.1#HandlerOptions
> 
> > // The first argument is a list of currently open groups that contain the
> > // Attr. It must not be retained or modified. ReplaceAttr is never called
> > // for Group attributes, only their contents.
> 
> Yep, I can check if this would work, because then is not a group but a single field!

Yeah groups can't be changed...this looks like a reason to not using grouping since we deliberately want to change the main object key name (the root httpRequest).

katcipis: > Thinking better, that's what I'd expect in the struct case as well.

yeah I tested with maps because I never used maps and was curious, it definitely works well with structs (for the JSON handler at least).

katcipis: > Not sure, I would need to test. But it seems to me, that it'll be `{"http_request": "{\"a\":\"b\"}"`, like a json inside of the string, not as different fields.

No it won't, I already use that a lot on our services with structs + I just tested with maps. It handles structured data well, it won't marshal it as a single huge string.

Here is an example with structs: https://go.dev/play/p/r2KCl6H4wPy

katcipis: Not behaving this way would make structured logging remarkably hard, makes perfect sense to handle well maps/structs and leverage their structures on the log entry.

katcipis: > I had a use-case on search that I wanted to know what was the log level set. Maybe an approach is being able to get those information. Which is the log level/format, then I could dynamically set the right field. Just an idea

I don't get the problem/solution here (why do you need to know the level and why changes on level would change fields that are set :thinking: ...I get it for the format, but that is handled by having a customized handler etc).

guilherme-santos: Yep, than seems to be a way. I'll try to do those changes and see how it goes

katcipis: It only prints everything as a single string on the TextHandler. If you want a nice output when using the TextHandler then it seems grouping is the only way to go (matches what I saw on the docs too). But I would not optimize for that, the output with the TextHandler is nice enough, and the JSON one works perfectly (for indexing/searching since it preserves the structure/fields).

https://go.dev/play/p/243w0A47rfr

slog/slog.go
guilherme-santos: Not really, but I also thought about this.

The `groups` means open groups... So, when the fields I'm interested in exists, the only open group available will be `httpRequest`, if I have another group inside of `httpRequest`, then `len(groups) == 2`. 

It's not all groups that exists!

katcipis: Ah OK, the groups is the whole path in the hierarchy of groups, so you are looking for the root group "httpRequest". I find this grouping thing kinda messy TBH...for what I checked playing around with the lib seems like it is only useful when you use the TextHandler..it is a way to group thins that the text handler understands and uses to serialize the log entries. But for JSON Handler it is redundant... a map/struct already does grouping perfectly fine. Unless you want to create a logger with a group and then pass the logger to some other function that then won't know it is logging inside a group...but I don't see a clear use case for that yet...I usually just add new fields to provide context..not groups...hmm

slog/slog.go
```diff
@@ -114,6 +119,32 @@ func Configure(cfg Config) error {
 	return nil
 }
 
+// Customize the http request fields
+// More: https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#HttpRequest
+func convertHTTPRequest(orig slog.Value) (string, slog.Value) {
+	var attrs []slog.Attr
+	origMap, _ := orig.Any().(map[string]any)
```
katcipis, line 126: Can't you just manipulate origMap, copying field X to Y like:

```go
origMap["requestMethod"] = origMap["method"]
delete(origMap, "method")
```

I imagine the mutations on this underlying map will be reflected on the given slog.Value no ? (maps are very side effecty =P...and I doubt that Any() returns a deep copy).

katcipis: Not completely against doing a full copy...specially if it is just per request..should not be a huge problem....it just seems like a no copy solution should be feasible here.

tracing/tracing.go
```diff
@@ -37,14 +37,25 @@ func InstrumentHTTP(h http.Handler) http.Handler {
 		}
 		ctx = slog.NewContext(ctx, log)
 
-		log.Debug("handling request", "url", req.URL, "method", req.Method)
+		httpReq := map[string]any{
```
katcipis, line 40: struct should also work no ? maybe that would be more clear/safe since it will be clear which fields are which type etc (there is no dynamic nature here AFAIK, only optional fields that can be ommited etc).

slog/slog.go
guilherme-santos: > Can't you just manipulate origMap, copying field X to Y like:
Yes, I started doing that... but indeed happen what you mention: we have 2 lines the "handling" and the "handled" reusing the same map, so if I delete one it affects the other. I could create different maps to workaround that.

I'm not really sure which approach is less problematic. If you're ok with copying I can keep it.

tracing/tracing.go
guilherme-santos: The problem is, if it's a struct defined inside of tracing package, I'll not be able to read it on the slog package.

slog/slog.go
katcipis: we can start with copying and if it is a problem we re-assess...I was just wondering why it would not be possible..because it seems possible =P

tracing/tracing.go
```diff
@@ -94,6 +108,11 @@ func (r *responseWriter) WriteHeader(statusCode int) {
 	r.ResponseWriter.WriteHeader(statusCode)
 }
 
+func (r *responseWriter) Write(b []byte) (int, error) {
+	r.contentLength += int64(len(b))
+	return r.ResponseWriter.Write(b)
```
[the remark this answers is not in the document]
guilherme-santos: Fixed!

tracing/tracing.go
katcipis: it doesn't convert the fields defined in the struct to a map internally ? I kinda expected that..but never tested since this is quite advanced.... hmm maybe it keeps the original struct and the JSON handler itself transforms it...so yeah it would not be available as a map...and then reflection is used on the JSON handler...yeah makes sense, that is why the TextHandler only prints the struct itself (probably using String()/same mechanism of the fmt.Print) and the JSON handler does something else.

All good then, disregard =P

katcipis: Although there seems to be an implicit protocol between the tracing pkg and the slog pkg..not sure keeping this implicit/dynamic is the best way to handle this...but maybe it is...hmm

slog/slog.go
katcipis, line 126:
```suggestion
	origMap, ok := orig.Any().(map[string]any)
	if !ok {
	    return orig
	}
```

Rationale: If for some reason someone added "http_request" = "some string" now we are going to remove the field since we return an empty value. We should keep the original field if it doesn't conform to what we expect.

tracing/tracing.go
guilherme-santos: Also a point!

slog/slog.go
katcipis: Actually my code is not enough the fix the whole problem...but I do believe that if someone is using the key "http_request" with some other type that is not a map..it should work as expected and no changes should be done on the field in that case, or else the lib behaves in a very odd way, converting valid data into empty with a different key name (the behavior may be quite surprising, if someone is not using the tracing pkg...or using a new version of it that changed something etc, it should do the translation only if the protocol seems to be right..at least the types etc).

guilherme-santos: Good catch! Fixed!

tracing/tracing_test.go
```diff
@@ -47,9 +50,12 @@ func TestIntrumentedHTTPHandler(t *testing.T) {
...
 	}
 
-	gotStatus := res.Result().StatusCode
-	if gotStatus != wantStatus {
-		t.Fatalf("got %v; want %v", gotStatus, wantStatus)
+	res := w.Result()
+	if got := res.StatusCode; got != wantStatus {
+		t.Fatalf("got status %v; want %v", got, wantStatus)
+	}
+	if got := w.Body.String(); got != wantBody {
```
katcipis, line 57: nice !

slog/slog.go
```diff
@@ -95,11 +95,16 @@ func Configure(cfg Config) error {
...
+				return a
 			}
-			if a.Key == slog.MessageKey {
+			switch a.Key {
+			case slog.LevelKey:
+				a.Key = "severity"
+			case slog.MessageKey:
 				a.Key = "message"
+			case "http_request":
+				a.Key, a.Value = convertHTTPRequest(a.Key, a.Value)
```
katcipis, line 107: not a huge deal..just thinking why the API is not just transforming the attribute, so it would be:

```
a = convertHTTPRequest(a)
```

Kinda like the signature of the ReplaceAttr function, it gets a slot.Attr and returns a transformed slog.Attr. But definitely not a big deal..or even a small deal I suppose...just wondering when I was re-reading the code :laughing:

tracing/tracing.go
katcipis: yeah..changes here will break there..silently....and adding new things will always be coordinated anyway (or it won't work properly for gcloud formatting)...hmm, we can assess this further later...really not sure about this

katcipis (approved):

PR #36, opened 2024-01-09
feat: add metadata to subscription messages

katcipis:
The idea is to get access to this metadata from Google PubSub: https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage

I don't think we will use this for metadata ourselves, we usually just model our own metadata inside the envelope, but Google metadata will be there so it seems that having at least a way to read metadata (but no way to define it) is enough. Not sure it is the best design..but I'm also not sure this will work to get this metadata, documentation is not super clear on this (if it doesn't work we can just remove it).

If this works then we can also add some tests, but first need to do some manual integration testing to see if this does what I think it does =P.

### DIFF
--- a/event/event.go
+++ b/event/event.go
@@ -23,13 +23,14 @@ type (
 
 	// Envelope represents the structure of all data that wraps all events.
 	Envelope[T any] struct {
-		// Metadata is publisher specific metadata. Including metadata that may be added by specific
+		TraceID string `json:"trace_id"`
+		OrgID   string `json:"organization_id"`
+		Name    string `json:"name"`
+		Event   T      `json:"event"`
+
+		// metadata is publisher specific metadata. Including metadata that may be added by specific
 		// pubsub brokers like Google Cloud PubSub.
-		Metadata map[string]string
-		TraceID  string `json:"trace_id"`
-		OrgID    string `json:"organization_id"`
-		Name     string `json:"name"`
-		Event    T      `json:"event"`
+		metadata map[string]string
 	}
 
 	// Subscription is a subscription that received only specific types of events
@@ -152,7 +153,7 @@ func (s *Subscription[T]) Serve(handler Handler[T]) error {
 			return fmt.Errorf("event name doesn't match %q: event: %v", s.name, msg)
 		}
 
-		event.Metadata = msg.Metadata()
+		event.metadata = msg.Metadata()
 
 		ctx = tracing.CtxWithTraceID(ctx, event.TraceID)
 		ctx = tracing.CtxWithOrgID(ctx, event.OrgID)
@@ -229,4 +230,9 @@ func (m Message) Metadata() map[string]string {
 // String representation of the message.
 func (m Message) String() string {
 	return string(m.body)
+}
+
+// Metadata returns all metadata associated with this event envelope.
+func (e Envelope[T]) Metadata() map[string]string {
+	return e.metadata
 }
--- a/event/event.go
+++ b/event/event.go
@@ -23,10 +23,13 @@ type (
 
 	// Envelope represents the structure of all data that wraps all events.
 	Envelope[T any] struct {
-		TraceID string `json:"trace_id"`
-		OrgID   string `json:"organization_id"`
-		Name    string `json:"name"`
-		Event   T      `json:"event"`
+		// Metadata is publisher specific metadata. Including metadata that may be added by specific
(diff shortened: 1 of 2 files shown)


lfritz (approved):
Right now you need a `MessageSubscription` to get the metadata, right? I would be nice if it also worked with a `Subscription`.

Maybe we can add another method to `Subscription`, something like this

    type EnvelopeHandler[T any] func(context.Context, Envelope[T]) error
    func (s *Subscription[T]) ServeEnvelope(handler EnvelopeHandler[T]) error

guilherme-santos (approved):
Nice, is there any useful information that we can get from there?

katcipis:
> Right now you need a `MessageSubscription` to get the metadata, right? I would be nice if it also worked with a `Subscription`.
> 
> Maybe we can add another method to `Subscription`, something like this
> 
> ```
> type EnvelopeHandler[T any] func(context.Context, Envelope[T]) error
> func (s *Subscription[T]) ServeEnvelope(handler EnvelopeHandler[T]) error 
> ```

Ah yeah I forgot we don't have access to the envelope today...hmmm. Well Im between something like that or a handler that accepts as parameters the event as we have today + the metadata (we don't have the envelope today because all important data is moved to the context). Hmm we can also add this metadata no the context...like we do with the trace ID and org ID..I usually don't like abusing the context ...but metadata is usually what it is used for so maybe it is a good idea here ?

katcipis:
> Nice, is there any useful information that we can get from there?

All data is here: https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage

- publishTime: Can be useful for ordering, since clocks on Google infra are usually quite good (True Time)
- messageId: Can be useful for dedup logic (it is the only way I know to be sure you are avoiding reprocessing the same thing VS the message only looks the same but it is not a duplicate, in the sense that it was published again).

lfritz:
> Ah yeah I forgot we don't have access to the envelope today...hmmm. Well Im between something like that or a handler that accepts as parameters the event as we have today + the metadata

That would also work.

> Hmm we can also add this metadata no the context...like we do with the trace ID and org ID..I usually don't like abusing the context ...but metadata is usually what it is used for so maybe it is a good idea here ?

Hm, adding data to the context means it's available not just to the function you call but also any functions it might call (as long as they accept a context). That makes sense for the trace ID and org ID but probably not for the metadata. It's really metadata for that specific Pub/Sub event.

katcipis:
> > Ah yeah I forgot we don't have access to the envelope today...hmmm. Well Im between something like that or a handler that accepts as parameters the event as we have today + the metadata
> 
> That would also work.
> 
> > Hmm we can also add this metadata no the context...like we do with the trace ID and org ID..I usually don't like abusing the context ...but metadata is usually what it is used for so maybe it is a good idea here ?
> 
> Hm, adding data to the context means it's available not just to the function you call but also any functions it might call (as long as they accept a context). That makes sense for the trace ID and org ID but probably not for the metadata. It's really metadata for that specific Pub/Sub event.

I disagree with the context not making sense for the reason you mentioned. The context that is created here is specific for a single pub/sub event, the trace ID and the organization ID are specific to a single Pub/Sub event. Storing the context and using it later in a different execution context would be wrong/confusing for example (the context is only valid for that specific event/handle call).

So if the argument is that only specific pub/sub event data should be there then IMHO that would be OK, the whole call stack handling an event is tied to that specific event, like if some function deeper on the event handling needs the publish time the fact that it will get from the context is not wrong IMHO (just as it uses the context for cancellation of that specific event handling)...just as we do with the organization ID...it is "request/event" specific data and the handler is handling that specific "request/event".

I'm more concerned with abusing context usage, overloading it with too much data and avoiding proper design of some concepts.

guilherme-santos:
Hey guys, I also feel that metadata shouldn't be in the context, metadata can be anything if I understood correctly we rven could have a attributes with a key, value, and add all this to the context seems to be a bit weird IMHO.

I think is ok to add to the ctx some of the information that we believe should be there (as trace and org id), but publish time this is something that we want to have in the ctx? I'd guess no, but it's a interesting information. So my opinion is should be available as proposed by the Metadata, but I don't think all metadata should be in the ctx, we can keep the cntx with what we have right now as it seems to be enough for our use-cases so far.

katcipis:
The metadata only gives us access to Google Cloud PubSub attriobutes...not to the actual metadata...so this won't solve our current problem x_x.

PR #37, opened 2024-01-11
feat: add event metadata and attributes

katcipis:


event/event.go
```diff
@@ -42,9 +42,27 @@ type (
...
 
+	// HandlerWithMetadata is responsible for handling events from a [Subscription] with the associated event metadata.
+	// The context passed to the handler will have all common metadata also passed to calls to [Handler] and extra metadata
+	// that is more specific as defined on [Metadata].
+	HandlerWithMetadata[T any] func(context.Context, T, Metadata) error
+
 	// Message represents a raw message received on a subscription.
 	Message struct {
-		body []byte
+		Body     []byte
```
katcipis, line 52: Since slices are a nightmare of mutability I think having a method added too much complexity without bringing too much benefits =/ (Im excited about avoiding mutability...but in Go it is not very easy to do that...)

event/event.go
```diff
@@ -199,23 +257,18 @@ func (r *MessageSubscription) Serve(handler MessageHandler) error {
 	}
 }
 
+func getMetadata(msg *pubsub.Message) (string, time.Time) {
+	// This is the only way to get broker specific metadata
+	// For now we only support Google Cloud.
```
katcipis, line 262: I tested this manually....so just trust me ? :laughing:

event/event.go
```diff
@@ -42,9 +43,26 @@ type (
 	// by using [slog.FromCtx].
 	Handler[T any] func(context.Context, T) error
 
+	// HandlerWithMetadata is responsible for handling events from a [Subscription] with the associated event metadata.
+	// The context passed to the handler will have all common metadata also passed to calls to [Handler] and extra metadata
+	// that is more specific as defined on [Metadata].
```
lfritz, line 48: Is that correct? I don't see where the extra metadata is added to the context.

lfritz (approved):

event/event.go
katcipis: It is..but it is confusing. I'm referencing the trace ID and the org ID as metadata...because they are :laughing: ...but yeah we have two sets of metadata...the general ones and the too specific ones (that you pointed out before). Hmm need to explain this better here.

katcipis: Tried to improve the docs =P

lfritz (approved):

PR #40, opened 2024-01-25
Add justfile and use it for CI

lfritz:
I also added a vulncheck CI target.

### DIFF
--- a/.github/workflows/vulncheck.yaml
+++ b/.github/workflows/vulncheck.yaml
@@ -1,6 +1,8 @@
-name: CI
+name: Vulnerability check
 
-on: [push]
+on:
+  schedule:
+    - cron: '0 7 * * 1'  # Weekly on Mondays at 7:00 AM
 
 jobs:
   test:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -1,53 +1,24 @@
...
-
   test:
-    name: Test
-
     runs-on: ubuntu-22.04
-
-    strategy:
-      matrix:
-        go:
-          - "1.21"
```
katcipis, line 32: I usually guarantee compatibility with older Go versions on my libs if possible, but I don't think this is a problem here =P

katcipis (approved):

.github/workflows/ci.yaml
lfritz: Oh that's why you had the matrix setup... yeah, in our case testing with older Go versions is probably excessive.

katcipis: yeah it was...it wasn't just because Im a crazy motherfucker :rofl: . Some of my libs work even with Go 1.16...so I keep testing =P...but here it is unnecessary indeed

katcipis: So I would just add 1.22 on the matrix when it was released, and so on (coincidentally now it requires at least 1.21).

PR #41, opened 2024-02-14
feat: add xhttp client retrier

katcipis:
Introduce the xhttp package. x may stand for:

* eXperimental
* eXtended
* EXTREME !!!
* XGuiga (the best option of course)

Take your pick :-). The idea is to avoid name clashing with stdlib packages, but extend them with things that are useful for us and more specific.

The main design idea of the retrier is to be a drop-in replacement for the stdlib http.Client (but for now we only implement Do, since it is what we use most of the time AFAIK).

Basic usage is:

```go
	client := xhttp.NewRetrierClient(&http.Client{})
        client.Do(req)
```

The same proxying can be done later for metrics/caching/etc.


xhttp/fake_client_test.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+// FakeClient allows to fake interactions with an [xhttp.Client].
+type FakeClient struct {
+	requests  []*http.Request
+	responses []response
+	mutex     sync.Mutex
+}
+
+func NewFakeClient() *FakeClient {
```
katcipis, line 16: Code that injects the xhttp.Client interface could use this FakeClient to test interactions with other services (investiganting requests + injecting responses).

katcipis: If this makes sense we can just publish it with the package :-)

xhttp/client.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+		return nil, err
+	}
+
+	if res.StatusCode == http.StatusInternalServerError ||
+		res.StatusCode == http.StatusServiceUnavailable ||
+		res.StatusCode == http.StatusTooManyRequests {
+		// Maybe add handling for Retry-After header, so far this seems to be enough
+		r.sleep(ctx, sleepPeriod)
+		return r.do(ctx, req, requestBody, sleepPeriod*2)
```
guilherme-santos, line 120: Will we retry forever? Also maybe the status could be customized, for instance in the API you're integrating Conflict should be retried as well.

guilherme-santos (approved):
Nice 👏🏻

xhttp/client.go
katcipis: I was not sure if we can generalize Conflict to always retry...but if you think so we can add it. Regarding re-trying forever, no, it is controlled by the context added on the request (how you normally would create a timeout when making a request using http.Client). If no context is defined, then yeah it is forever (which is the same behavior of the http.Client since it can literally hang forever waiting for an answer).

We can add an extra max retries thing... so far I didn't use it, I just always define a context with a timeout and use it when sending requests (and that controls the worse case of how long I will wait for a response).

katcipis: In general I think about retrying more in terms of how long I'm willing to wait than how much times did I retry already, because there is no correlation with how much times you tried already with being safe to give up, so I just try until I have time to do it.

guilherme-santos: I don't think we should add Conflict here, most cases we shouldn't retry. My point was about be configurable.

Regarding the ctx, I understand, but not sure if time will always be a good metric to stop. Nevertheless, if you don't need for your use-case we can always extend later.

katcipis: For Conflict Im not sure if in general just retrying with the exact same request always works... so I was less prone to define it here...maybe it could be an option ? (it could be an option that keeps adding status code that you also want to retry, and retries on any of them), like:

```go
xhttp.NewRetrierClient(c, xhttp.RetrierWithRetryOn(http.StatusConflict), xhttp.RetrierWithRetryOn(http.StatusSomethingElse))
```

guilherme-santos: Yep! That was what I had in mind! 👍🏻 
Feel free to do later if someone needs it.

katcipis: > I don't think we should add Conflict here, most cases we shouldn't retry. My point was about be configurable.
> 
> Regarding the ctx, I understand, but not sure if time will always be a good metric to stop. Nevertheless, if you don't need for your use-case we can always extend later.

Ah yeah I missed that you already mentioned customization, yeah I think that makes sense and can be done :-).
Regarding the max retries, what is the scenario that you think it would be interesting ? we already have an exponential backoff, so even if we retry multiple times the sleep gets longer and longer, what would be the max retries that you would think is a good idea to give up ? Every time I'm integrating I always want to retry for as long as I can, until I have no more time, or else..what is a good number of retries ? 3 ? 10 ?

If a service is down for too long and we want to avoid sending requests to it I would say the solution is a circuit breaker, but I surivived so long just with exponential backoffs that Im not quite sure we need them.

katcipis: But Im not against adding it as a parameter, and by default we don't use it (and tries until context expires)

guilherme-santos: I just feel that retrying by time will also include time used by the server, let's say that your request takes 2, 3 secs, how many retries fit there, don't really know, may fit a lot (which can be a problem as we're causing loading in a service that is already down), and may be too less. I feel that I retry X times more clear of my intentions comparing with I retry for X secs that again it may be too much or too little.

katcipis: For the problem of causing load to a service I believe a circuit breaker is the proper solution, having max retries only alleviates the problem (barely, if you retry a lot, and if you retry too little you risk failing unnecessarily in some cases).

Then regarding feeling better with retrying N times, I usually feel better retrying for N period of time =P. Retrying N times has the advantage that it may send less requests to a service that is down, but that is not a huge problem since we have an exponential backoff...we will not just hammer down like crazy a downed service (+ the proper solution is the circuit breaker if this is a huge problem). But it has the disadvantage of, lets say you configured 5 max retries...what if the sixth time it would work ? and the timeout you had configured to perform the job was 30 secs and you still had 29 seconds available ? why fail ? in this case you just failed the request unnecessarily, you had plenty of time to finish it and the server just got back up. When thinking in those terms, I never know how much retries will be the best...5 ? what if 6th  would work ? 10 ? what if the 11th would work ? and so goes on =P.

So by default I think no max retries and having a backoff protocol makes more sense in order to optimize to working instead of failing unnecessarily.

katcipis: Also, for a time based definition + exponential backoff  we can know the worse case scenario of how much requests will happen based on the sleep period. For a service that is down and fails immediately. If the context is configured to wait for 10 seconds, the worse case scenario is (for the default 1 second of min sleep period and considering the request fails immediately):

- first request
- 1 sec 
- second request
- 2 sec
- third request
- 4 sec
- fourth request
- 8 sec

It will never make more than 4 requests (because of the exponential backoff + min sleep period). We never have an unbounded number of requests and it is always possible to compute the worse case scenario. It is only unbounded if the context is unbounded (no time limit/cancellation), but then that is the caller option.

guilherme-santos: Yep, it seems fine, then!

xhttp/fake_client_test.go
lfritz: I think that's a good idea. Users of xhttp might want to use it in their own tests.

lfritz (approved):

PR #43, opened 2024-02-18
feat: add retry on custom status codes

katcipis:
Also add some logging support and retry on deadline exceeded errors

### DIFF
--- a/xhttp/client.go
+++ b/xhttp/client.go
@@ -3,6 +3,7 @@ package xhttp
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -144,7 +145,8 @@ func (r *retrierClient) do(ctx context.Context, req *http.Request, requestBody [
 		// For connections reset...same problem:
 		// - https://github.com/golang/go/blob/d0dc93c8e1a5be4e0a44b7f8ecb0cb1417de50ce/src/net/http/transport_test.go#L2207
 		// We need to go for some suffix matches.
-		if strings.Contains(err.Error(), "http2: server sent GOAWAY and closed the connection") ||
+		if errors.Is(err, context.DeadlineExceeded) ||
+			strings.Contains(err.Error(), "http2: server sent GOAWAY and closed the connection") ||
 			strings.HasSuffix(err.Error(), ": connection reset by peer") {
 			r.logger.Debug("xhttp.Client: retrying request with error", "error", err, "sleep_period", sleepPeriod)
 			r.sleep(ctx, sleepPeriod)
--- a/xhttp/client.go
+++ b/xhttp/client.go
@@ -62,6 +62,11 @@ func NewRetrierClient(c Client, options ...RetrierOption) Client {
 		client:    c,
 		sleep:     defaultSleep,
 		minPeriod: time.Second,
+		retryStatusCodes: map[int]struct{}{
+			http.StatusInternalServerError: {},
+			http.StatusServiceUnavailable:  {},
+			http.StatusTooManyRequests:     {},
+		},
 	}
 	for _, option := range options {
 		option(r)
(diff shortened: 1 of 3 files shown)


xhttp/client.go
```diff
@@ -55,13 +55,26 @@ func RetrierWithRequestTimeout(timeout time.Duration) RetrierOption {
 	}
 }
 
+// RetrierWithStatus will configure the retrier to retry when this specific status code is received.
+// Multiple of this option can be passed to configure multiple different status code.
+func RetrierWithStatus(status int) RetrierOption {
```
lfritz, line 60: This only lets us add status codes (not remove them) so it'll always retry on the default ones, right? Maybe it would be good have a way to override the defaults... we could define the method like this so you can replace all the codes:

    func RetrierWithStatus(statuses map[int]struct{}) RetrierOption

lfritz (approved):

xhttp/client.go
katcipis: I would just change it to be a slice of status in that case (no need to expose the internal map/set structure that is a little odd because Go does not have sets :laughing: ). I was wondering if we would ever really do that, the current default status codes are quite limited to things that usually we really want to retry (500 and 503). Do you already have a use case in mind ? or else we could use it for a while and then revisit this.

katcipis: Maybe we could add later a "RetrierWithStatuses" if necessary (that then is a set that replaces the internal set). If we only have the option that replaces everything then every time I need a new status I need to remember all the basic status code that I need to retry (like 503) and if I forget them it is broken x_x.

lfritz: Yeah, I see the value of having sensible defaults... The one status I'm wondering about is `http.StatusTooManyRequests`. We might want to handle that at a higher level.

katcipis: > Yeah, I see the value of having sensible defaults... The one status I'm wondering about is `http.StatusTooManyRequests`. We might want to handle that at a higher level.

hmm I see, I do get this status a lot from LLM services and in general I just retry after some time exactly as I do with Service unavailable (both can have a header Retry-After if Im not wrong, which could be used here too, to make it more precise maybe ?)...knowing where you are in a window of status too many requests can be tricky, most APIs don't provide precise information..so your only option is to keep retrying anyway or just failing.

What I mean is, what would be the logic you would do differently ? (besides handling a Retry-After header ?) If we already have this use case..an alternative would be to remove "http.StatusTooManyRequests" from default handling, if it is that specific.

lfritz: So far we don't have any special handling for 429 status in the new connectors. I think we do in some of the connectors in feedback-river; not sure how we're handling it there.

katcipis: Yeah...so far Im designing things around my own experience but Im aware it doesn't cover all use cases...but would be good to have a clear use case in mind...WDYT ? we remove it as a default status code to be on the safe side ? I don't mind defining it manually on my services.

lfritz: Yup, that's a good solution for now.

lfritz (dismissed):

xhttp/client.go
```diff
@@ -103,18 +145,19 @@ func (r *retrierClient) do(ctx context.Context, req *http.Request, requestBody [
 		// For connections reset...same problem:
 		// - https://github.com/golang/go/blob/d0dc93c8e1a5be4e0a44b7f8ecb0cb1417de50ce/src/net/http/transport_test.go#L2207
 		// We need to go for some suffix matches.
-		if strings.Contains(err.Error(), "http2: server sent GOAWAY and closed the connection") ||
+		if errors.Is(err, context.DeadlineExceeded) ||
+			strings.Contains(err.Error(), "http2: server sent GOAWAY and closed the connection") ||
 			strings.HasSuffix(err.Error(), ": connection reset by peer") {
-
+			r.logger.Debug("xhttp.Client: retrying request with error", "error", err, "sleep_period", sleepPeriod)
```
guilherme-santos, line 151: We have the log in the ctx, right? Wouldn't be better to use it instead. Something that I would be missing here is the trace_id for example, getting from the ctx we have all the information we need otherwise the log will be lose

guilherme-santos: I'm just not sure if we need to avoid the log dependency, other packages inside of the lib already depend on it, and the lib (even though is public is focused on our use-cases, don't remember if we decide that or not), so I'd be ok to have the dependency. Just feel that will be really easy to miss the log as we would not have the trace-id to match. I was also ok to not having logging TBH.

The hook seems a interesting approach, maybe even the retry codes could be one hook, or kind of middleware that chained decide if retry or not, giving a lot of control for the caller. Or maybe too much things 😅

lfritz: I would tend to just have it depend on our slog package to keep it simple.

lfritz (approved):

PR #50, opened 2024-03-05
fix: context cancellation with per request timeouts

katcipis:
When we create per requests contexts we were cancelling the contexts before when we returning from Do,
this caused responses to fail when they are being read, since the response reading is part of the
original request context (and then can timeout/get cancelled).

### DIFF
--- a/xhttp/client_test.go
+++ b/xhttp/client_test.go
@@ -230,6 +230,9 @@ func TestRetrierPerRequestTimeoutWontCancelContext(t *testing.T) {
 	if req.Context().Err() == nil {
 		t.Fatal("want request context to be cancelled after closing response body")
 	}
+	if ctx.Err() != nil {
+		t.Fatal("parent context should not be cancelled after closing response body")
+	}
 }
 
 func TestRetrierExponentialBackoff(t *testing.T) {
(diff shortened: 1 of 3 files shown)


xhttp/client_test.go
```diff
@@ -119,12 +116,120 @@ func TestRetrierPerRequestTryTimeout(t *testing.T) {
...
+		t.Fatalf("request context is cancelled: %v", req.Context().Err())
+	}
+
+	gotBody, err := io.ReadAll(res.Body)
+	if err != nil {
+		t.Fatalf("reading response body: %v", err)
+	}
+	assertEqual(t, string(gotBody), wantResponseBody)
+
+	// after the response body is closed then context created for the specific request should be cancelled
```
katcipis, line 222:
```suggestion
	// after the response body is closed then the context created for the specific request should be cancelled
```

xhttp/client_test.go
```diff
@@ -119,12 +116,123 @@ func TestRetrierPerRequestTryTimeout(t *testing.T) {
...
+		t.Fatalf("got %d requests; want 1", len(requests))
+	}
+
+	// Guarantee that the context created for the request was cancelled
+	req := requests[0]
+	if req.Context().Err() == nil {
+		t.Fatal("want request context to be cancelled after response error")
+	}
+	if ctx.Err() != nil {
+		t.Fatal("parent context should not be cancelled after response error")
```
katcipis, line 164:
```suggestion
		t.Fatal("want request context to be cancelled after error response")
	}
	if ctx.Err() != nil {
		t.Fatal("parent context should not be cancelled after error response")
```

lfritz (approved):

PR #55, opened 2024-04-19
event: better log message for panic recovery

lfritz:
When event.MessageSubscription catches a panic in the message handler, we end up with this confusing error in the logs:

    !ERROR:json: unsupported type: func(func(inte

This change should result in a more useful log.

### DIFF
--- a/event/event.go
+++ b/event/event.go
@@ -266,7 +266,7 @@ func (r *MessageSubscription) Serve(handler MessageHandler) error {
 					buf = buf[:runtime.Stack(buf, false)]
 					slog.Error("panic: message subscription: handling message",
 						"error", err,
-						"message", msg.Body,
+						"message_body", msg.Body,
 						"metadata", msg.Metadata,
 						"stack_trace", string(buf))
 					if gocloudMsg.Nackable() {
--- a/event/event.go
+++ b/event/event.go
@@ -256,9 +256,12 @@ func (r *MessageSubscription) Serve(handler MessageHandler) error {
 					const size = 64 << 10
 					buf := make([]byte, size)
 					buf = buf[:runtime.Stack(buf, false)]
+					messageID, publishTime := getMetadata(msg)
 					slog.Error("panic: message subscription: handling message",
 						"error", err,
-						"message_body", msg.Body,
+						"message_body", string(msg.Body),
+						"message_id", messageID,
+						"publish_time", publishTime,
 						"stack_trace", string(buf))
 					if msg.Nackable() {
 						msg.Nack()
--- a/event/event.go
+++ b/event/event.go
@@ -237,7 +237,7 @@ func (r *MessageSubscription) Serve(handler MessageHandler) error {
 	semaphore := make(chan struct{}, r.maxConcurrency)
 	for {
 		semaphore <- struct{}{}
-		msg, err := r.sub.Receive(context.Background())
+		gocloudMsg, err := r.sub.Receive(context.Background())
 		if err != nil {
 			// From: https://pkg.go.dev/gocloud.dev@v0.30.0/pubsub#example-Subscription.Receive-Concurrent
 			// Errors from Receive indicate that Receive will no longer succeed.
@@ -248,7 +248,15 @@ func (r *MessageSubscription) Serve(handler MessageHandler) error {
 				<-semaphore
 			}()
 
-			id, publishedTime := getMetadata(msg)
+			id, publishedTime := getMetadata(gocloudMsg)
+			msg := Message{
+				Body: gocloudMsg.Body,
+				Metadata: Metadata{
+					ID:            id,
+					PublishedTime: publishedTime,
+					Attributes:    gocloudMsg.Metadata,
+				},
+			}
 
 			defer func() {
 				if err := recover(); err != nil {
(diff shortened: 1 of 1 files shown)


event/event.go
```diff
@@ -256,7 +256,10 @@ func (r *MessageSubscription) Serve(handler MessageHandler) error {
 					const size = 64 << 10
 					buf := make([]byte, size)
 					buf = buf[:runtime.Stack(buf, false)]
-					slog.Error("panic: message subscription: handling message", "message", msg, "stack_trace", string(buf))
```
katcipis, line 259: Not getting why the old error was producing: 

"!ERROR:json: unsupported type: func(func(inte"

msg is a struct...and stack_trace is a big string... :confused:

event/event.go
```diff
@@ -256,7 +256,10 @@ func (r *MessageSubscription) Serve(handler MessageHandler) error {
...
+					slog.Error("panic: message subscription: handling message",
+						"error", err,
+						"message_body", msg.Body,
```
katcipis, line 261: Why not print the message with metadata too ?

event/event.go
katcipis: Ah it is related to "message" being used by google cloud logging as a special field

event/event.go
lfritz: When I tried it using just `msg` as the value, it printed the message as a byte array. Example:

```
2024/04/19 12:56:40 ERROR panic: message subscription: handling message message="&{LoggableID:msg #0 Body:[123 34 116 114 97 99 101 95 105 100 34 58 34 34 44 34 111 114 103 97 110 105 122 97 116 105 111 110 95 105 100 34 58 34 34 44 34 110 97 109 101 34 58 34 104 101 108 108 111 34 44 34 101 118 101 110 116 34 58 123 34 116 101 120 116 34 58 34 66 111 110 106 111 117 114 34 125 125] Metadata:map[] BeforeSend:<nil> AfterSend:<nil> asFunc:0x851600 ack:0x82a600 nackable:true mu:{state:0 sema:0} isAcked:false}" stack_trace="goroutine 27 [running]:\ngithub.com/birdie-ai/golibs/event.(*MessageSubscription).Serve.func1.2()\n\t/home/levin/code/golibs/event/event.go:258 +0x59\npanic({0x8cdf00?, 0xdc73c0?})\n\t/usr/local/go/src/runtime/panic.go:770 +0x132\nmain.handler({0xc0003b23c0?, 0x4e?}, {{0xc000126245?, 0x0?}})\n\t/home/levin/code/foo/main.go:48 +0x7f\ngithub.com/birdie-ai/golibs/event.(*Subscription[...]).Serve.func1()\n\t/home/levin/code/golibs/event/event.go:169 +0x18a\ngithub.com/birdie-ai/golibs/event.(*Subscription[...]).Serve.SampledMessageHandler.func2()\n\t/home/levin/code/golibs/event/metrics.go:21 +0x11c\ngithub.com/birdie-ai/golibs/event.(*MessageSubscription).Serve.func1()\n\t/home/levin/code/golibs/event/event.go:266 +0x1a4\ncreated by github.com/birdie-ai/golibs/event.(*MessageSubscription).Serve in goroutine 9\n\t/home/levin/code/golibs/event/event.go:246 +0x47\n"
```

That's not super-useful :smile:

katcipis: ah damn..true...fucking byte arrays. Maybe log both on separate fields but forcing the body to be string

katcipis (approved):

event/event.go
```diff
@@ -256,7 +256,13 @@ func (r *MessageSubscription) Serve(handler MessageHandler) error {
 					const size = 64 << 10
 					buf := make([]byte, size)
 					buf = buf[:runtime.Stack(buf, false)]
-					slog.Error("panic: message subscription: handling message", "message", msg, "stack_trace", string(buf))
+					messageID, publishTime := getMetadata(msg)
+					slog.Error("panic: message subscription: handling message",
+						"error", err,
+						"message_body", string(msg.Body),
+						"message_id", messageID,
+						"publish_time", publishTime,
```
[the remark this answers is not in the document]
lfritz: I got a bit confused about what you meant by message metadata... what `getMetadata` returns is completely different from what `msg.Metadata` contains, right?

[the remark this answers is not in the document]
katcipis: not completely different, I think it is more like a subset, I need to remember what is this function used for :laughing: (this is confusing because some brokers talk about metadata and attributes, in some of them the concepts are inverted...so abstracting this is hard..but there is usually both things).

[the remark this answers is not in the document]
katcipis: Ah yeah looked at the function, yes that is a more low level thing that is used to get the metadata from the underlying google thing. It is used here so you don't need to call it again: https://github.com/birdie-ai/golibs/blob/5cb693a7a0a035e413dba3e15ab4bdc0c3220f00/event/event.go#L251

But I also got confused.... we have two messages, the google one and the one we actually handle: https://github.com/birdie-ai/golibs/blob/5cb693a7a0a035e413dba3e15ab4bdc0c3220f00/event/event.go#L272

I thought "msg" would be ours, but it is the google one. My previous code was even wronger than I imagined, sorry I built this error handling and never tested it...should have done some manual testing at least x_x. Now I get why the error message was extra odd....I was printing the library type...not our message type (that is much simpler).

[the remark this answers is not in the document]
katcipis: What we could do here is before the recover create our own msg, our type, and then use that on the logging for the recover. Maybe call the other one "rawMsg" and then ours "msg".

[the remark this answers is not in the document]
katcipis: Sorry Im giving you so much work, I did this in a hurry and now you are paying the price for my incompetence :disappointed:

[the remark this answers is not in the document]
lfritz: Done

event/event.go
```diff
@@ -248,36 +248,41 @@ func (r *MessageSubscription) Serve(handler MessageHandler) error {
...
 			}()
 
-			id, publishedTime := getMetadata(msg)
+			id, publishedTime := getMetadata(gocloudMsg)
+			msg := Message{
+				Body: gocloudMsg.Body,
+				Metadata: Metadata{
+					ID:            id,
+					PublishedTime: publishedTime,
+					Attributes:    gocloudMsg.Metadata,
```
katcipis, line 257: This is the confusing part...gocloud calls metadata what google calls attributes, pubsub has both...I ended up following the google design.

event/event.go
```diff
@@ -248,36 +248,41 @@ func (r *MessageSubscription) Serve(handler MessageHandler) error {
...
 					// 64KB, if it is good enough for Go's stdlib it is good enough for us :-)
 					const size = 64 << 10
 					buf := make([]byte, size)
 					buf = buf[:runtime.Stack(buf, false)]
-					slog.Error("panic: message subscription: handling message", "message", msg, "stack_trace", string(buf))
-					if msg.Nackable() {
-						msg.Nack()
+					slog.Error("panic: message subscription: handling message",
+						"error", err,
+						"message", msg.Body,
```
katcipis, line 269:
```suggestion
						"message_body", msg.Body,
```

Or else we will lose the original message we passed on the log, the "panic: message subscription: handling message", that being the "main" message is nice since it is easier to filter.

katcipis (approved):
YESSS !!!
Thanks :heart:

PR #56, opened 2024-05-14
feat: add callback for requests on xhttp.RetrierClient

katcipis:
This is a first building block for adding some client side metrics (also logging when desired).

### DIFF
--- a/xhttp/client.go
+++ b/xhttp/client.go
@@ -109,9 +109,9 @@ func (r *retrierClient) do(ctx context.Context, req *http.Request, requestBody [
 
 	log := slog.FromCtx(ctx).With("request_url", req.URL)
 
+	start := time.Now()
 	res, err := r.client.Do(req)
-	// TODO: handle errors/responses and calculate proper elapsed
-	r.onRequestDone(req, res, err, 0)
+	r.onRequestDone(req, res, err, time.Since(start))
 	if err != nil {
 		cancel()
 
--- a/xhttp/client.go
+++ b/xhttp/client.go
@@ -23,6 +23,13 @@ type (
 	}
 	// RetrierOption is used to configure retrier clients created with [NewRetrierClient].
 	RetrierOption func(*retrierClient)
+
+	// RetrierOnRequestDoneFunc is the callback called when using [RetrierWithOnRequestDone].
+	// The [http.Request] is the original http request that just finished.
+	// The [time.Duration] is how long the http request took to be finished.
+	// This is called for every request that is done, including retries.
+	// Any changes made on the given [http.Request] may be reflected on retry behavior.
+	RetrierOnRequestDoneFunc func(req *http.Request, res *http.Response, elapsed time.Duration)
 )
 
 const (
@@ -37,10 +44,11 @@ const (
 // The returned [Client] will automatically retry failed requests.
 func NewRetrierClient(c Client, options ...RetrierOption) Client {
 	r := &retrierClient{
-		client:    c,
-		sleep:     defaultSleep,
-		minPeriod: DefaultMinSleepPeriod,
-		maxPeriod: DefaultMaxSleepPeriod,
+		client:        c,
+		sleep:         defaultSleep,
+		minPeriod:     DefaultMinSleepPeriod,
+		maxPeriod:     DefaultMaxSleepPeriod,
+		onRequestDone: defaultOnRequestDone,
 		retryStatusCodes: map[int]struct{}{
 			http.StatusInternalServerError: {},
 			http.StatusServiceUnavailable:  {},
@@ -61,6 +69,7 @@ type (
 		checkResponse    bool
 		sleep            func(context.Context, time.Duration)
 		retryStatusCodes map[int]struct{}
+		onRequestDone    RetrierOnRequestDoneFunc
 	}
 	readerCloserCanceller struct {
 		io.ReadCloser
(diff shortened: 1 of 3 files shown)


xhttp/client.go
```diff
@@ -23,6 +23,14 @@ type (
 	}
 	// RetrierOption is used to configure retrier clients created with [NewRetrierClient].
 	RetrierOption func(*retrierClient)
+
+	// RetrierOnRequestDoneFunc is the callback called when using [RetrierWithOnRequestDone].
+	// The [*http.Request] is the original http request that just finished.
+	// The [*http.Response] is the response of the request or nil if the request failed with an error (then the error will be non-nil).
```
i4ki, line 29:
```suggestion
	// The [*http.Response] is the response of the request or nil if the request failed with an error.
```
I think this goes without saying.

xhttp/client.go
```diff
@@ -99,7 +109,9 @@ func (r *retrierClient) do(ctx context.Context, req *http.Request, requestBody [
 
 	log := slog.FromCtx(ctx).With("request_url", req.URL)
 
+	start := time.Now()
 	res, err := r.client.Do(req)
+	r.onRequestDone(req, res, err, time.Since(start))
```
i4ki, line 114: where's the guarantee for the invariant that `res == nil` imply `err != nil` and vice versa? unless the `r.client` is not Go's HTTP client, in that case, ignore.

i4ki (approved):
I dont have much context but well done :-)
Nice docs. The test logic looks okay but depends on a testing framework I'm unable to check right now.

I miss reviewing your code...

I left a small suggestion and a question, feel free to ignore.

xhttp/client.go
katcipis: hmm I assumed this is guaranteed by the underlying http.Do that we wrap, usually when you have an error you should ignore the returned value..but indeed not always. I was re-reading the [docs](https://pkg.go.dev/net/http#Client.Do)...and actually they don't explicitly say that the res will be nil, there is only an invariant related to the error being nil and the body being always non-nil. Maybe it is better to just document that we will pass on the callback what is returned by the Do method instead of trying to duplicate/guarantee invariants, this is supposed to be very lightweight and just give you a view on what was returned by Do/underlying request.

xhttp/client.go
katcipis: I will improve those docs and how invariants are documented, seems like just redirecting what is returned by Do makes more sense than trying to re-specify invariants.

katcipis:
> I dont have much context but well done :-) Nice docs. The test logic looks okay but depends on a testing framework I'm unable to check right now.

hmm there isn't any frameworks here..only Go's testing package + we do have a mock for http.Client that is used for testing here + also used when testing our services, but I consider it just a library =P (nothing very big/frameworky...I hope).

> 
> I miss reviewing your code...

I miss your reviews :heart: 

> 
> I left a small suggestion and a question, feel free to ignore.

I would never ignore your review.

xhttp/client.go
katcipis: > for context: https://cs.opensource.google/go/go/+/refs/tags/go1.22.3:src/net/http/client.go;l=710-719

Just now saw the link =P. Yeah I brainfarted because of the invariant of "when the error is nil the body is always non-nil", I think I extrapolated some other stuff from that.

In the end I'm not sure we should enforce a lot of invariants, it all depends on the underlying http.Client...usually it is the Go one...but it could be a different one...we don't enforce that.

i4ki:
oh I didn't mean that kind of framework haha but yes I meant the `xhttptest` package. I just saw it now, it's quite simple.

xhttp/client.go
i4ki: > In the end I'm not sure we should enforce a lot of invariants, it all depends on the underlying http.Client...usually it is the Go one...but it could be a different one...we don't enforce that.

makes sense.

katcipis: Yeah the docs also say that actually, so just read the docs I suppose :rofl: 

"A non-nil Response with a non-nil error only occurs when CheckRedirect fails, and even then the returned [Response.Body] is already closed."

katcipis: It is quite complex...in the end..this whole protocol (at least to me..given my obvious limitations :rofl: )...because every time you have a valid response you are supposed to close the body..and body won't be nil, and usually when there is an error there is no response...but when there is an error of a specific kind then there will be a response..but in that case you don't need to close the body :laughing: (and it is safe to ignore it...as it is usual on Go's error handling).

katcipis:
> oh I didn't mean that kind of framework haha but yes I meant the `xhttptest` package. I just saw it now, it's quite simple.

HELL YEAH IT IS !!! :rofl: (I was slightly offended by framework =P)

xhttp/client.go
i4ki: yep... as usual, because they need to maintain compatibility with Go 1, they make things more complicated than needed. For almost everything else we take this for granted, that if `err != nil` then actual value/response is nil/empty.
The cognitive load of maintaining software just increases...
I see the same problem with the `io.Reader` interface ... some many cases to handle because of past bad decisions...

i4ki: no problem with bad decisions... they happen but improving is important.

i4ki (approved):
LGTM

lfritz (approved):

xhttp/client.go
katcipis: io.Reader and now the http.Client.Do method always make me wonder how code is not enough to specify these protocols, you need out of band natural language specifications to document the protocol (and it is harder to detect bugs).

PR #58, opened 2024-05-17
Add xtime package with a type for a time range

lfritz:
We use time range in the enricher, the scheduler, and the connectors, so why not add them here.

### DIFF
--- a/xtime/range.go
+++ b/xtime/range.go
@@ -3,31 +3,31 @@ package xtime
 
 import "time"
 
-// Range defines a time range that contains time t if From <= t < To.
+// Range defines a time range that contains time t if Start <= t < End.
 type Range struct {
-	From, To time.Time
+	Start, End time.Time
 }
 
 // Contains returns true if [t] is within the range.
 func (r Range) Contains(t time.Time) bool {
-	return !t.Before(r.From) && t.Before(r.To)
+	return !t.Before(r.Start) && t.Before(r.End)
 }
 
 // Duration returns the duration of the time range.
 func (r Range) Duration() time.Duration {
-	return r.To.Sub(r.From)
+	return r.End.Sub(r.Start)
 }
 
 // Split returns a list of time ranges of at most [max] length that together make up [r].
 func (r Range) Split(max time.Duration) []Range {
 	var result []Range
-	for max != 0 && r.To.Sub(r.From) > max {
-		next := r.From.Add(max)
+	for max != 0 && r.End.Sub(r.Start) > max {
+		next := r.Start.Add(max)
 		result = append(result, Range{
-			From: r.From,
-			To:   next,
+			Start: r.Start,
+			End:   next,
 		})
-		r.From = next
+		r.Start = next
 	}
 	result = append(result, r)
 	return result
--- a/xtime/range.go
+++ b/xtime/range.go
@@ -1,0 +1,34 @@
+// Package xtime extends Go's time. So far it contains an implementation of a time range.
+package xtime
+
+import "time"
+
+// Range defines a time range that contains time t if From <= t < To.
+type Range struct {
+	From, To time.Time
+}
+
+// Contains returns true if [t] is within the range.
+func (r Range) Contains(t time.Time) bool {
+	return !t.Before(r.From) && t.Before(r.To)
+}
+
+// Duration returns the duration of the time range.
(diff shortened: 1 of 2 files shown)


xtime/range_test.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+}
+
+func TestRangeDuration(t *testing.T) {
+	cases := []struct {
+		from, to time.Time
+		want     time.Duration
+	}{
+		{tm(1, 0), tm(1, 0), time.Duration(0)},
+		{tm(1, 0), tm(2, 0), 1 * time.Hour},
```
katcipis, line 43:
```suggestion
		{tm(1, 0), tm(2, 0), time.Hour},
```

Or disregard, not sure 1 * adds much here readability wise.

xtime/range_test.go
```diff
@@ -0,0 +1,99 @@
...
+		{
+			tm(1, 0),
+			tm(2, 0),
+			2 * time.Hour,
+			[]xtime.Range{{tm(1, 0), tm(2, 0)}},
+		},
+		{
+			tm(1, 0),
+			tm(2, 0),
+			1 * time.Hour,
```
katcipis, line 70:
```suggestion
			time.Hour,
```

You probably just like this....so ignore me being annoying :rofl:

katcipis (dismissed):
Damn you took it literally :rofl:, great stuff (chef's kiss)

xtime/range_test.go
lfritz: I find it a bit more readable -- in English you'd say "it takes one hour" not "it takes hour". Maybe that's just me though.

xtime/range.go
```diff
@@ -0,0 +1,34 @@
+// Package xtime extends Go's time. So far it contains an implementation of a time range.
+package xtime
+
+import "time"
+
+// Range defines a time range that contains time t if From <= t < To.
+type Range struct {
+	From, To time.Time
```
katcipis, line 8: Now...the eternal struggle in my mind...should this be from/to...or start/end ? Currently Im working with start/end on the enricher =P.

katcipis: When I think about ranges in a general purpose way it matches more with start/end

xtime/range_test.go
katcipis: I imagined you had some reasoning like that :laughing: ...I find it odd because you are saying more than it is necessary, it feels like I need to spend some extra cycles to be sure that isn't something else going on =P (because it is a no-op/identity operation.. so my mind always go to "why ?" instead of just reading as english).

xtime/range.go
lfritz: In the ingestion we're mixing both :sweat_smile:

I don't have a preference, but we should probably agree on a standard.

katcipis: Not sure... :rofl: (sometimes I use from/to...but usually on the CLI because then the CLI itself is more specific and it seems to read better there)

xtime/range_test.go
lfritz: OK, if others find it confusing then I'll change it

katcipis: it could be just me :laughing: ...feel free to ignore it

xtime/range.go
katcipis: but on APIs I went with start/end..for the themes....hmmm. Im slightly more prone towards start/end.

katcipis: WDYT @guilherme-santos ?

xtime/range_test.go
guilherme-santos: I'm quite ok with 1 as well actually :)

katcipis: See, it is just me being an awkward motherfucker :rofl:

xtime/range.go
guilherme-santos: On search I also have a range, but it's more like a operator Greater, GreaterOrEqual and so on. But I do have some cases where I call Start/End for time, and From/To to other things (like id for instance).

katcipis: OK Ill vote for start/end :rofl: (but Im OK with from/to).

guilherme-santos: I think for time start/end is quite nice! +1 for start/end!

katcipis (approved):

PR #60, opened 2024-05-23
xhttp.Client: follow Retry-After header

lfritz:
I want to start using the retrierClient to retry on status 429, but it seems like then we should also use the Retry-After header.

### DIFF
--- a/xhttp/client.go
+++ b/xhttp/client.go
@@ -167,7 +167,7 @@ func (r *retrierClient) do(ctx context.Context, req *http.Request, requestBody [
 		switch {
 		case err != nil:
 			log.Warn(fmt.Sprintf("xhttp.Client: %v", err))
-		case requestedDuration > minRetryAfterDuration:
+		case requestedDuration >= minRetryAfterDuration:
 			log.Debug("xhttp.Client: following Retry-After header", "duration", requestedDuration)
 			sleepPeriod = requestedDuration
 		case !requestedTime.IsZero():
--- a/xhttp/client.go
+++ b/xhttp/client.go
@@ -161,18 +161,22 @@ func (r *retrierClient) do(ctx context.Context, req *http.Request, requestBody [
 		r.onRetry(req, res, err)
 
 		// handle Retry-After header
+		const minSleepPeriod = time.Second
 		retryAfter := res.Header.Get("Retry-After")
 		requestedDuration, requestedTime, err := ParseRetryAfter(retryAfter)
 		switch {
 		case err != nil:
 			log.Warn(fmt.Sprintf("xhttp.Client: %v", err))
-		case requestedDuration != 0:
+		case requestedDuration > minSleepPeriod:
 			log.Debug("xhttp.Client: following Retry-After header", "duration", requestedDuration)
 			sleepPeriod = requestedDuration
 		case !requestedTime.IsZero():
-			sleepPeriod = time.Until(requestedTime)
-			log.Debug("xhttp.Client: following Retry-After header", "time", requestedTime,
-				"calculated_duration", requestedDuration)
+			calculatedDuration := time.Until(requestedTime)
+			if calculatedDuration > minSleepPeriod {
+				log.Debug("xhttp.Client: following Retry-After header", "time", requestedTime,
+					"calculated_duration", calculatedDuration)
+				sleepPeriod = calculatedDuration
+			}
 		}
 
 		r.sleep(ctx, sleepPeriod)
--- a/xhttp/client.go
+++ b/xhttp/client.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
@@ -158,7 +159,22 @@ func (r *retrierClient) do(ctx context.Context, req *http.Request, requestBody [
 		}
 		log.Debug("xhttp.Client: retrying request with error status code")
 		r.onRetry(req, res, err)
-		// Maybe add handling for Retry-After header, so far this seems to be enough
+
+		// handle Retry-After header
+		retryAfter := res.Header.Get("Retry-After")
+		requestedDuration, requestedTime, err := ParseRetryAfter(retryAfter)
+		switch {
+		case err != nil:
+			log.Warn(fmt.Sprintf("xhttp.Client: %v", err))
+		case requestedDuration != 0:
+			log.Debug("xhttp.Client: following Retry-After header", "duration", requestedDuration)
+			sleepPeriod = requestedDuration
+		case !requestedTime.IsZero():
+			sleepPeriod = time.Until(requestedTime)
+			log.Debug("xhttp.Client: following Retry-After header", "time", requestedTime,
+				"calculated_duration", requestedDuration)
+		}
+
 		r.sleep(ctx, sleepPeriod)
 		return r.do(ctx, req, requestBody, min(sleepPeriod*2, r.maxPeriod))
 	}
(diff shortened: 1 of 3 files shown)


xhttp/client_opts.go
```diff
@@ -51,7 +51,6 @@ func RetrierWithRespCheck() RetrierOption {
 }
 
 // RetrierWithSleep configures the sleep function used to sleep between retries, usually used for testing.
-// But can be used as a way to measure how much retries happened since this is called before each retry.
```
lfritz, line 54: This is irrelevant now that we have the callback functions, right?

katcipis: It is not irrelevant in the sense that it provides different information (the sleep times) and also it allows users to override sleeping behavior.

katcipis: ah if you mean the docs, yeah that part of the docs is misleading because we have better and more clear ways to do that now.

lfritz: Yeah, I just meant that one sentence in the docs.

xhttp/client.go
```diff
@@ -206,3 +222,17 @@ func defaultOnRequestDone(*http.Request, *http.Response, error, time.Duration) {
...
 }
+
+// ParseRetryAfter parses the Retry-After header in the response.
+func ParseRetryAfter(value string) (time.Duration, time.Time, error) {
+	if value == "" {
+		return 0, time.Time{}, nil
+	}
+	if intValue, err := strconv.Atoi(value); err == nil {
+		return time.Duration(intValue) * time.Second, time.Time{}, nil
+	}
```
katcipis, line 233:
```suggestion
	if seconds, err := strconv.Atoi(value); err == nil {
		return time.Duration(seconds) * time.Second, time.Time{}, nil
	}
```

xhttp/client.go
```diff
@@ -158,7 +159,22 @@ func (r *retrierClient) do(ctx context.Context, req *http.Request, requestBody [
...
+		retryAfter := res.Header.Get("Retry-After")
+		requestedDuration, requestedTime, err := ParseRetryAfter(retryAfter)
+		switch {
+		case err != nil:
+			log.Warn(fmt.Sprintf("xhttp.Client: %v", err))
+		case requestedDuration != 0:
+			log.Debug("xhttp.Client: following Retry-After header", "duration", requestedDuration)
+			sleepPeriod = requestedDuration
+		case !requestedTime.IsZero():
+			sleepPeriod = time.Until(requestedTime)
```
katcipis, line 173: What if the sleep period is negative ? (the requested time is on the past). I know it would be some problem/bug on the server side but that would cause us to start sending requests as fast as possible on each try. Actually it could also be a bug/issue with clock synchronization on our side, making our time.Now to be way in the future, etc. IMHO when for any reason this is smaller than 1 second (it is the smallest retry period on the header when you use the duration on the header) we should use the configured sleep period and assume there is something wrong with the headers/clocks (wall clocks are nasty... that is why usually the duration is a better idea..but the header support both).

xhttp/client_test.go
```diff
@@ -903,3 +903,26 @@ func assertEqual[T any](t *testing.T, got T, want T) {
...
+		tm    time.Time
+	}{
+		{value: ""},
+		{value: "123", d: 123 * time.Second},
+		{value: "Wed, 21 Oct 2015 07:28:00 GMT", tm: time.Date(2015, 10, 21, 7, 28, 0, 0, time.UTC)},
+	}
+	for _, c := range cases {
+		d, tm, err := xhttp.ParseRetryAfter(c.value)
+		if err != nil {
+			t.Errorf("ParseRetryAfter(%q) returned error: %v", c.value, err)
```
katcipis, line 920: maybe test some error situation, just be sure it doesn't panic etc :-)

xhttp/client.go
katcipis: Having said that...the same should be done for negative durations...it doesn't make sense but on the web you just expect the worse =P..in this case the error is solely on the server side...but would be better to handle it gracefully.

lfritz: I added a `minSleepPeriod` to avoid those cases

xhttp/client_test.go
lfritz: Done

xhttp/client.go
```diff
@@ -158,7 +159,26 @@ func (r *retrierClient) do(ctx context.Context, req *http.Request, requestBody [
 		}
 		log.Debug("xhttp.Client: retrying request with error status code")
 		r.onRetry(req, res, err)
-		// Maybe add handling for Retry-After header, so far this seems to be enough
+
+		// handle Retry-After header
+		const minSleepPeriod = time.Second
```
katcipis, line 164:
```suggestion
		const minRetryAfterDuration = time.Second
```

It could be mistaken with some sort of min/max sleep duration we have on the retrier (we have a max one..but no min...but it got me thinking it could be interpreted as that).

xhttp/client.go
```diff
@@ -158,7 +159,26 @@ func (r *retrierClient) do(ctx context.Context, req *http.Request, requestBody [
...
+		retryAfter := res.Header.Get("Retry-After")
+		requestedDuration, requestedTime, err := ParseRetryAfter(retryAfter)
+		switch {
+		case err != nil:
+			log.Warn(fmt.Sprintf("xhttp.Client: %v", err))
+		case requestedDuration > minSleepPeriod:
```
katcipis, line 170:
```suggestion
		case requestedDuration >= minSleepPeriod:
```

1 second is valid, in this case 1 second would be ignored. Overriding the sleep function could help testing this kind of sneaky logic if necessary.

xhttp/client.go
```diff
@@ -158,7 +159,26 @@ func (r *retrierClient) do(ctx context.Context, req *http.Request, requestBody [
...
+			log.Debug("xhttp.Client: following Retry-After header", "duration", requestedDuration)
+			sleepPeriod = requestedDuration
+		case !requestedTime.IsZero():
+			calculatedDuration := time.Until(requestedTime)
+			if calculatedDuration > minSleepPeriod {
```
katcipis, line 175:
```suggestion
			if calculatedDuration >= minSleepPeriod {
```

xhttp/client_test.go
```diff
@@ -903,3 +903,40 @@ func assertEqual[T any](t *testing.T, got T, want T) {
...
+		"1.5",
+		"Wed, 32 Oct 2015 07:28:00 GMT",
+	}
+	for _, c := range cases {
+		_, _, err := xhttp.ParseRetryAfter(c)
+		if err == nil {
+			t.Errorf("xhttp.ParseRetryAfter(%q) did not return error", c)
+		}
+	}
+}
```
katcipis, line 942: :heart:

xhttp/client.go
```diff
@@ -158,7 +159,26 @@ func (r *retrierClient) do(ctx context.Context, req *http.Request, requestBody [
...
+		requestedDuration, requestedTime, err := ParseRetryAfter(retryAfter)
+		switch {
+		case err != nil:
+			log.Warn(fmt.Sprintf("xhttp.Client: %v", err))
+		case requestedDuration >= minRetryAfterDuration:
+			log.Debug("xhttp.Client: following Retry-After header", "duration", requestedDuration)
+			sleepPeriod = requestedDuration
+		case !requestedTime.IsZero():
+			calculatedDuration := time.Until(requestedTime)
+			if calculatedDuration > minRetryAfterDuration {
```
katcipis, line 175:
```suggestion
			if calculatedDuration >= minRetryAfterDuration {
```

lfritz: Oops, I thought I changed both

katcipis (approved):

PR #64, opened 2024-06-19
Subscription: optionally ack messages immediately

lfritz:
Not sure if this is the right approach, so only a draft for now


lfritz:
Closing this, it's probably not the way to go

PR #70, opened 2024-08-20
feat:tracing: add fine grained control over request stats

katcipis:
Some services handle way too much requests (dozens of millions per day). Logging each single request on the
INFO level is starting to generate considerable costs (and we rarely need those logs..if ever). In order to be
able to change this behavior Im introducing a new type and a callback for the request stats. The current default
behavior will be maintained.

### DIFF
--- a/tracing/tracing.go
+++ b/tracing/tracing.go
@@ -10,6 +10,18 @@ import (
 	"github.com/google/uuid"
 )
 
+// RequestStats contains stats for a completed request.
+type RequestStats struct {
+	Method       string `json:"method,omitempty"`
+	URL          string `json:"url,omitempty"`
+	RequestSize  int64  `json:"request_size,omitempty"`
+	UserAgent    string `json:"user_agent,omitempty"`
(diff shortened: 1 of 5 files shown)


tracing/tracing.go
```diff
@@ -10,10 +10,34 @@ import (
...
+// RequestStats contains stats for a completed request.
+type RequestStats struct {
+	Method       string `json:"method,omitempty"`
+	URL          string `json:"url,omitempty"`
+	RequestSize  int64  `json:"request_size,omitempty"`
+	UserAgent    string `json:"user_agent,omitempty"`
+	Protocol     string `json:"protocol,omitempty"`
+	StatusCode   int    `json:"status_code,omitempty"`
+	ResponseSize int    `json:"response_size,omitempty"`
+	Elapsed      string `json:"elapsed,omitempty"`
```
katcipis, line 22: Same as url.URL...but for time.Duration.

lfritz (approved):

PR #74, opened 2024-10-16
feat: add retry on "use of closed network connection" err

katcipis:


xhttp/client_test.go
```diff
@@ -125,14 +125,14 @@ func TestRetrierPerRequestTryTimeout(t *testing.T) {
 
 	// make sure that while we cancel created internal contexts we dont accidentally cancel the parent context
 	if ctx.Err() != nil {
-		t.Fatalf("want original context to be valid but got cancelled: %v", ctx.Err())
```
lfritz, line 128: Just for context: both of these are correct (https://en.wiktionary.org/wiki/cancel#Verb)

lfritz (approved):

xhttp/client_test.go
katcipis: Ah my spell checking thing was always saying it must be "canceled" not "cancelled" ....I was a little confused since I was pretty sure cancelled was correct...but I just assumed I was being stupid (as usual :laughing: )

PR #80, opened 2025-02-05
feat: add xerrgroup

katcipis:
Very basic extension to errgroup. Have been using conc for this, but realized that what I want is very simple. Simpler than what they have there, the API has a lot of different types/constructors/error modes/cancellation/etc...I never remember which one I want, because what I want is pretty much errgroup but I want auto collection of results because Im lazy and I do this a lot.

It is a design choice to not allow zero values to be valid, I like being able to use the zero value but sadly I would not be able to support WithContext since it returns a pointer and we cant copy the underlying errgroup. If we drop WithContext it would be possible to make it work with zero values, but WithContext can be important in scenarios where you want to be able to cancel all subtasks if one of them failed.


go.mod
```diff
@@ -25,10 +26,9 @@ require (
 	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.51.0 // indirect
 	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.51.0 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
-	golang.org/x/net v0.25.0 // indirect
-	golang.org/x/sync v0.7.0 // indirect
-	golang.org/x/sys v0.20.0 // indirect
-	golang.org/x/text v0.15.0 // indirect
+	golang.org/x/net v0.33.0 // indirect
```
katcipis, line 29: unrelated, dependabot is being annoying about this.

matmarzolagomes (approved):
Cool!

PR #82, opened 2025-03-09
feat: add xjson

katcipis:
Some utilities around the json package. Helpers to decode stream/single values and dynamic objects handling.

### DIFF
--- a/xjson/xjson.go
+++ b/xjson/xjson.go
@@ -35,8 +35,7 @@ type (
 )
 
 // UnmarshalFile calls [Unmarshal] with the opened file (closing it afterwards) and returns the unmarshalled value.
-// If you need more details, like the data that was read when an unmarshalling error happened,
-// you can:
+// If you need more details, like the data that was read when an unmarshalling error happened, you can:
 //
 //	var errDetails UnmarshalError
 //	if errors.As(err, &errDetails) {
@@ -56,8 +55,7 @@ func UnmarshalFile[T any](path string) (T, error) {
 
 // Unmarshal calls [json.Unmarshal] after reading the given reader into memory
 // and returns the unmarshalled value.
-// If you need more details, like the data that was read when an unmarshalling error happened,
-// you can:
+// If you need more details, like the data that was read when an unmarshalling error happened, you can:
 //
 //	var errDetails UnmarshalError
 //	if errors.As(err, &errDetails) {


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/birdie-ai/golibs
 
-go 1.22
+go 1.23
```
gustavom2998, line 3: Question - how do you go about updating this?

Keep the minimum version you can that will still keep the code working?

xjson/xjson.go
```diff
@@ -0,0 +1,155 @@
...
+	v, ok := anyV.(T)
+	if !ok {
+		return z, false, fmt.Errorf("value at path %q: expected to have type %T but has %T", path, z, anyV)
+	}
+
+	return v, true, nil
+}
+
+func traverse(o Obj, path string) (string, Obj, error) {
+	parsedPath := strings.Split(path, ".")
```
gustavom2998, line 136: Dumb question maybe - won't this break if the JSON key includes "."


I've seens objects like `{"a.b":123}`

Maybe you just don't want to treat this. This happened to me once (and i wasn't expecting it) - it was a surprise 😆

xjson/xjson.go
```diff
@@ -0,0 +1,155 @@
...
+
+// NewDecoder creates a new decoder for type T.
+func NewDecoder[T any](r io.Reader) *Decoder[T] {
+	return &Decoder[T]{json.NewDecoder(r), nil}
+}
+
+// All returns a single-use iterator for the stream.
+func (d *Decoder[T]) All() iter.Seq[T] {
+	return func(yield func(v T) bool) {
+		for d.d.More() {
```
gustavom2998, line 84: What exactly are we iterating over here? 

Would we be getting the JSON via some HTTP stream or something and the decode it as we receive it?

Curious around what we could use this for.

go.mod
katcipis: Yes. Specially for libraries it should represent the minimum required version, so people are free to use any version they want that is compatible.

xjson/xjson.go
katcipis: I don't want to support this, none of the JSON I worked so far do this so doesn't seem like something we need to handle. If we have to then it would require:

- Some escaping
- Or being able to control the delimiter, but every delimiter you choose may break some other JSON anyway..but sed works like this...and it is nice

For now just not solving the problem.

katcipis: But yes...I'm aware it is possible :-)

xjson/xjson.go
katcipis: Streaming in general, usually done with JSONL: https://jsonlines.org/

PR #86, opened 2025-05-25
feat: add xhttp.NewJSONRequest

katcipis:
My idea is to add a layer as small as possible just to make it easier to handle JSON on HTTP request/responses. Usually errors are disjointed from responses (different fields), so I'm thinking on always creating a type that is the union of whatever you expect as response + what you would expect if it is an error. After checking the status code then you can validate each part of the parsed data, the error for an error status code and the valid response for a success. This allows for more sophisticated things to be built on top, like something that already checks the status code and hides HTTP entirely, but not sure that is necessary/desirable once we have something that already makes it considerably less annoying to receive JSON (no more reading/error handling/parsing for every request).

The current design interoperates with anything that creates an [http.Request] and the [xhttp.Client] interface is implemented by Go's [http.Client] and our retrier (basically, everything just works :man_shrugging: ).

### DIFF
--- a/xhttp/request.go
+++ b/xhttp/request.go
@@ -1,7 +1,9 @@
 package xhttp
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"io"
 	"net/http"
 	"runtime/debug"
@@ -59,4 +61,14 @@ func NewRequestWithContext(ctx context.Context, method, url string, body io.Read
 	}
 	req.Header.Set("User-Agent", defaultUserAgent)
 	return req, nil
+}
+
+// NewJSONRequest is a wrapper that will marshall the given data as a JSON and call [NewRequestWithContext].
+// It makes it practical to create HTTP requests that send JSON bodies with detailed user agents (what we want most of the time).
+func NewJSONRequest(ctx context.Context, method, url string, data any) (*http.Request, error) {
+	body, err := json.Marshal(data)
+	if err != nil {
+		return nil, err
+	}
+	return NewRequestWithContext(ctx, method, url, bytes.NewReader(body))
 }
--- a/xhttp/xhttp.go
+++ b/xhttp/xhttp.go
@@ -1,0 +1,57 @@
+package xhttp
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+)
+
+type (
+	// Response is an extension of [http.Response] that contains parsed data
+	// instead of a response body.
+	Response[T any] struct {
+		*http.Response
+		Obj T
+	}
+	// ResponseErr is the error returned by [Do] if parsing the response body fails.
+	ResponseErr struct {
+		Err  error
+		Body []byte
+	}
+)
+
+// Do calls [Client.Do] and unmarshalls the HTTP response as a JSON of type [T].
+// The returned [Response] embeds the original [http.Response] with the addition
+// of the [Response.Obj] field that holds the parsed response.
+//
+// The original [http.Response.Body] will always be read and closed, the caller should ignore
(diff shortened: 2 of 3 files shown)


xhttp/request.go
```diff
@@ -60,3 +62,13 @@ func NewRequestWithContext(ctx context.Context, method, url string, body io.Read
 	req.Header.Set("User-Agent", defaultUserAgent)
 	return req, nil
 }
+
+// NewJSONRequest is a wrapper that will marshall the given data as a JSON and call [NewRequestWithContext].
+// It makes it practical to create HTTP requests that send JSON bodies with detailed user agents (what we want most of the time).
+func NewJSONRequest(ctx context.Context, method, url string, data any) (*http.Request, error) {
```
katcipis, line 68: Adding WithContext seemed like too much here, in general we always pass a context to ensure cancellation/timeouts.

xhttp/xhttp.go
```diff
@@ -0,0 +1,57 @@
...
+// Do calls [Client.Do] and unmarshalls the HTTP response as a JSON of type [T].
+// The returned [Response] embeds the original [http.Response] with the addition
+// of the [Response.Obj] field that holds the parsed response.
+//
+// The original [http.Response.Body] will always be read and closed, the caller should ignore
+// this field and use [Response.Obj] to access the parsed response or use errors.As
+// to check details in the case of an error (eg. debugging malformed JSON).
+//
+// If the response is not valid JSON an error of type [ResponseErr] is returned.
+func Do[T any](c Client, req *http.Request) (*Response[T], error) {
```
katcipis, line 34: Thinking on leaving it just Do....we use JSON for 99% of HTTP interactions + there is no other reason that I can think off to have another layer on top of Do except this. But could also be DoJSON if someone feels like it would be better.

xhttp/xhttp.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		return nil, errors.Join(err, v.Body.Close())
+	}
+	if err := v.Body.Close(); err != nil {
+		return nil, fmt.Errorf("xhttp: closing response body: %w", err)
+	}
+
+	var parsed T
+	if err := json.Unmarshal(body, &parsed); err != nil {
+		return nil, ResponseErr{err, body}
```
katcipis, line 50: Maybe also add the original [http.Response] here so the caller can inspect more things ?

xhttp/xhttp.go
i4ki: we talked about this and I think the following would be useless but a nice addition haha:

```go
resp, _ := xhttp.Do[string](c, req)
if resp.Obj == "some string" {}
```

```go
resp, _ := xhttp.Do[[]byte](c, req)
if resp.Obj[0] == 10 {}
```
```
resp, _ := xhttp.Do[MyType](c, req)
if resp.Obj.Value == "some value" {}
```
etc

where `string` and `[]byte` would be special cases and all the rest handled as JSON.

but please ignore, JSON is used 100.00% of the time!

i4ki (approved):
lgtm

xhttp/request.go
katcipis: I thought the same :heart:

xhttp/xhttp.go
katcipis: Yeah that is a nice idea. But on the more pragmatic sense, lately I have been focusing more and more in a form of lightweight JSON RPC in my head. In the end HTTP/JSON won and everything uses it...and in general I like text protocols...and JSON is not great..but I suppose it could be worse ? :rofl: 

Of course sometimes you do need a binary protocol (like embeddings, we use JSON but it is painfully verbose for huge sets of numbers)...and for that I have no solution at all :laughing: (by not liking protobuf =P).

PR #89, opened 2025-06-05
feat: add xhttp.Response.RawBody

katcipis:
Should be useful mostly for debugging, but can help when an endpoint is sending you some JSON but it is different from what you expected (so parsing succeeds but information is not there).

Also renaming Obj to Value, Obj assumed too much about the structure.

### DIFF
--- a/xhttp/xhttp.go
+++ b/xhttp/xhttp.go
@@ -13,10 +13,10 @@ type (
 	// instead of a response body.
 	Response[T any] struct {
 		*http.Response
-		// RawObj is the raw response from where Obj was parsed (useful mostly for debugging).
-		RawObj []byte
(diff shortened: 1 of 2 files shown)


xhttp/xhttp.go
```diff
@@ -13,6 +13,9 @@ type (
 	// instead of a response body.
 	Response[T any] struct {
 		*http.Response
+		// RawObj is the raw response from where Obj was parsed (useful mostly for debugging).
+		RawObj []byte
```
katcipis, line 17: Not sure about this name, open to suggestions (didn't want to shadow the Response.Body).

i4ki (approved):
LGTM

PR #92, opened 2025-06-18
feat: add elapsed on xhttp.Response

katcipis:


xhttp/xhttp.go
```diff
@@ -52,9 +59,9 @@ func Do[T any](c Client, req *http.Request) (*Response[T], error) {
 
 	var parsed T
 	if err := json.Unmarshal(body, &parsed); err != nil {
-		return nil, ResponseParseErr{err, body, v.StatusCode}
+		return nil, ResponseParseErr{err, body, v.StatusCode, elapsed}
 	}
-	return &Response[T]{Response: v, RawBody: body, Value: parsed}, nil
+	return &Response[T]{v, body, parsed, elapsed}, nil
```
i4ki, line 64: fine by me but Go police dislikes instantiation in this way :)

i4ki (approved):

xhttp/xhttp.go
katcipis: I wasn't aware there was a strong dislike, I usually avoid it if the types on the struct are the same and it is easier to mess things up.

i4ki (approved):

PR #93, opened 2025-06-24
feat: add xerrors.Tag

katcipis:


i4ki (approved):
lgtm but as `t.err` and `t.tag` are private, for debugging purposes, maybe makes sense to have a `.Tag() error` that returns the tag?
then I can assert for `interface Tag() error` and use it.
But I don't know if actually needed, we can add later if I feel this urge many times.

katcipis:
> lgtm but as `t.err` and `t.tag` are private and it's invisible (not shown in `.Error()`), for debugging purposes, maybe makes sense to have a `.Tag() error` that returns the tag? then I can assert for `interface Tag() error` and use it. But I don't know if actually needed, we can add later if we feel this urge many times.

Wouldn't that be covered by errors.As to retrieve the tag ?

i4ki:
yeah that would work for checking if error tag is what you expect but not for getting whatever tag is in the error. I think it's okay. Feel free to merge.

katcipis:
> yeah that would work for checking if error tag is what you expect but not for getting whatever tag is in the error. I think it's okay. Feel free to merge.

hmm yeah I would need a more concrete use case/problem to get my head around it. Like you wanted to omit the error message but now you want to retrieve it, but it is not about retrieving/checking a specific error type with more details (which you would use errors.As) :thinking:

PR #95, opened 2025-07-16
feat: add google ordered publisher

katcipis:


cmd/testing/pubsub/README.md
```diff
@@ -0,0 +1,27 @@
+# PubSub Testing
+
+Very simple tooling to check ordered event delivery.
+Export variables:
+
+```sh
+export GCP_PROJECT=<project>
```
i4ki, line 7:
```suggestion
export GOOGLE_PROJECT=<project>
```
or maybe change the code :)

cmd/testing/pubsub/README.md
```diff
@@ -0,0 +1,27 @@
...
+go run . subscriber
+```
+
+And then:
+
+```sh
+go run . publisher
+```
+
+Resources won't be automatically deleted (except subscription that have an expiration policy of an hour), remember to delete them.
```
i4ki, line 27:
```suggestion
Resources won't be automatically deleted (except subscription that have an expiration policy of one day), remember to delete them.
```
one day is the minimum expiration.

cmd/testing/pubsub/main.go
```diff
@@ -0,0 +1,136 @@
...
+	projectID := os.Getenv("GOOGLE_PROJECT")
+	if projectID == "" {
+		log.Fatal("missing env var: GOOGLE_PROJECT")
+	}
+	topicName := os.Getenv("TOPIC_NAME")
+	if topicName == "" {
+		log.Fatal("missing env var: TOPIC_NAME")
+	}
+
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt)
```
i4ki, line 41: as `<-ctx.Done()` is not selected, this prevents user from terminating the subscriber.

cmd/testing/pubsub/main.go
```diff
@@ -0,0 +1,136 @@
...
+		log.Fatalf("creating subscription: %v", err)
+	}
+
+	log.Printf("starting handler with concurrency=%d", totalPartitions)
+	sub, err := event.NewOrderedGoogleSub[Event](ctx, projectID, topicName, eventName, totalPartitions)
+	panicerr(err)
+
+	err = sub.Serve(ctx, func(_ context.Context, event Event) error {
+		fmt.Printf("partition %q: count %d\n", event.PartitionID, event.Count)
+		time.Sleep(5 * time.Second)
```
i4ki, line 82: is this needed? I ran before checking the code and I thought it was unusable slow haha

event/google.go
```diff
@@ -0,0 +1,142 @@
...
+	})
+	_, err = res.Get(ctx)
+	sample(p.eventName, len(encBody), err)
+
+	return xerrors.Tag(err, ErrUnrecoverable)
+}
+
+// Resume must be called for the given orderingKey after a Publish call with the same
+// orderingKey failed and the error is [ErrUnrecoverable].
+func (p *OrderedGooglePublisher[T]) Resume(_ context.Context, orderingKey string) error {
```
i4ki, line 72: why returning an error?

i4ki: okay, found you comment in the test: https://github.com/birdie-ai/golibs/pull/95/files#r2213263572

xerrors/xerrors.go
```diff
@@ -16,6 +16,9 @@ import "errors"
 // Calls to [errors.As] and [errors.Is] will be dispatched to the tag first
 // and then fallback to the original error if they don't match the tag.
 func Tag(err, tag error) error {
+	if err == nil {
```
i4ki, line 19: won't this hide a programming error?

event/google.go
```diff
@@ -0,0 +1,142 @@
...
+
+	sample := publishSampler()
+	res := p.topic.Publish(ctx, &pubsub.Message{
+		OrderingKey: orderingKey,
+		Data:        encBody,
+	})
+	_, err = res.Get(ctx)
+	sample(p.eventName, len(encBody), err)
+
+	return xerrors.Tag(err, ErrUnrecoverable)
```
i4ki, line 67: noticed that you allowed calling it with `nil` but I wonder if this is a good practice.
Usually, the idea of hiding error handling does not Go well.

i4ki (approved):
lgtm but added minor comments.

cmd/testing/pubsub/main.go
katcipis: well define "needed", I wanted to be able to notice that the 10 goroutines received the events concurrently, the sleep allowed me to confirm that (10 prints happen at once), but I can remove it now ;-)

event/google.go
katcipis: Well...errors.Join on the stdlib does this =P...not sure it is a terrible idea...but no strong opinions...but then I would add an immediate panic there to avoid failing later in odd ways.

xerrors/xerrors.go
i4ki: I meant when the user has a call to `xerrors.Tag(err, SomeTagErr)` where `err` is nil by mistake. This way it will return `nil` and "work" instead of panicking later when printing the error (AFAICS it would panic in `err.Error()`).

IMO adding a panic is better because this is a misuse of the API.

event/google.go
i4ki: yeah I see some cases where it can be useful, coalescing non-nil values is a good one but not sure in general.
In this case it's a bit odd, because if supported I imagine people will do this often and then it's a new trend of bubbling up errors with optional wrappers without err handling =)

cmd/testing/pubsub/main.go
i4ki: yeah, good to keep it.

PR #109, opened 2025-11-11
feat: introduce dml.

i4ki:
Introduces a data manipulation language to capture change request events inside search.

### DIFF
--- a/bdml/doc.go
+++ b/bdml/doc.go
@@ -1,2 +1,0 @@
-// Package bdml handles encoding/decoding of the Birdie Data Manipulation Language.
-package bdml
--- a/dml/README.md
+++ b/dml/README.md
@@ -1,11 +1,12 @@
-# bdml: Birdie Data Manipulation Language
+# dml: Data Manipulation Language
 
-This package provides the parser and encoder for the Birdie Data Manipulation
-Language (bdml).
+This package provides the parser and encoder for the Data Manipulation Language (dml) used
+as write interface for the search services.
 
-## What is bdml?
+## What is dml?
 
-The `bdml` language exists for three reasons:
+DML stands for [Data Manipulation Language](https://en.wikipedia.org/wiki/Data_manipulation_language) and this `dml` exists
+primarily for three reasons:
 
 1. Having a common language that captures the data change semantics independent of the transport protocol used (REST, PubSub, etc).
 2. The ability to funnel data change events from different sources into a common language that allows for ease de-duplication and other optimizations.
--- a/dml/dml.go
+++ b/dml/dml.go
@@ -1,4 +1,4 @@
-package bdml
+package dml
 
 import "unique"
 
--- a/dml/dml_example_test.go
+++ b/dml/dml_example_test.go
@@ -1,23 +1,23 @@
-package bdml_test
+package dml_test
 
 import (
 	"fmt"
 	"unique"
 
-	"github.com/birdie-ai/golibs/bdml"
+	dml "github.com/birdie-ai/golibs/dml"
 )
 
 func ExampleStmt() {
-	stmt := bdml.Stmt{
-		Op:     bdml.SET,
+	stmt := dml.Stmt{
+		Op:     dml.SET,
 		Entity: unique.Make("mydata"),
-		Assign: bdml.Assign{
+		Assign: dml.Assign{
 			"kind.rating": 10,
 		},
-		Where: bdml.Clauses{
+		Where: dml.Clauses{
 			{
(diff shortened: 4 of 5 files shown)


bdml/README.md
```diff
@@ -0,0 +1,59 @@
...
+3. Separate higher level business logic (mostly data transformation) from lower level storage modifications.
+
+## Syntax
+
+Simplified eBNF:
+```
+PROGRAM = STMT+
+STMT = OP IDENT [ASSIGNLIST] 'WHERE' CLAUSE ';'
+OP = 'SET' | 'DELETE'
+ASSIGNLIST = NAME '=' VALUE [',' ASSIGNLIST ]+
```
matmarzolagomes: shouldn't it be *? 0 or more instead of 1 or more

bdml/bdml.go
```diff
@@ -0,0 +1,44 @@
+package bdml
+
+import "unique"
+
+type (
+	// Stmt is a single bdml statement.
+	Stmt struct {
+		Entity unique.Handle[string]
+		Op     OpKind
```
matmarzolagomes: just Op or Operation? sorry for nitpicking hahaha just giving alternatives but all good with naming so far tbh

i4ki: I'm choosing short/terse names because there will be thousands of test cases filling up these data structures by hand. Apart from that, `op` is a common short name for such thing.

bdml/README.md
```diff
@@ -0,0 +1,59 @@
...
+1. Having a common language that captures the data change semantics independent of the transport protocol used (REST, PubSub, etc).
+2. The ability to funnel data change events from different sources into a common language that allows for ease de-duplication and other optimizations.
+3. Separate higher level business logic (mostly data transformation) from lower level storage modifications.
+
+## Syntax
+
+Simplified eBNF:
+```
+PROGRAM = STMT+
+STMT = OP IDENT [ASSIGNLIST] 'WHERE' CLAUSE ';'
```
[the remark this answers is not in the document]
i4ki: it's optional because of `DELETE feedbacks WHERE id='abc'`

bdml/README.md
i4ki: yeah sorry, I just pasted here the draft version. 
Later I'm gonna make it a valid eBNF so we can also validate/test it in code.

bdml/bdml.go
```diff
@@ -0,0 +1,44 @@
+package bdml
+
+import "unique"
+
+type (
+	// Stmt is a single bdml statement.
+	Stmt struct {
```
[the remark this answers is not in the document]
matmarzolagomes: all good

bdml/README.md
```diff
@@ -0,0 +1,80 @@
+# bdml: Birdie Data Manipulation Language
```
gustavom2998:
```suggestion
# bdml: Birdie Data Manipulation Language
```

This indicates that all of birdie will use this language.

THis seems to be something especific to search, no?

gustavom2998: For example, if this is really Birdie wide, I'd expect all of our APIs to accept CRUD operations using this... Which isn't going to be the case.

bdml/bdml_example_test.go
```diff
@@ -0,0 +1,27 @@
+package bdml_test
+
+import (
+	"fmt"
+	"unique"
+
+	"github.com/birdie-ai/golibs/bdml"
+)
+
+func ExampleStmt() {
```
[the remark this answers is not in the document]
gustavom2998: Nice, thanks!

bdml/README.md
i4ki: > THis seems to be something especific to search, no?

We have the read DSL and now we are introducing a write DSL. This one is likely too low-level for several use cases but the idea is that soon any team can store data within search (beside the other entities). We are going to push up a lot of business logic, handling them in higher level services (like our preprocessor) and specialize some search services as an engine to write and read.

> For example, if this is really Birdie wide, I'd expect all of our APIs to accept CRUD operations using this..

Is it true that everything here must be used by all teams?

The idea to put this here was because it will be a library used by many services (not all) and it's gonna be quite small, then the project "golibs" fit the purpose, no? If this is a problem, let me know that I will ask for another repo specifically for this.

i4ki: hmm gotcha.
The "Search" in the name confuses with filtering/reading data, which is not the case. I can drop the "Birdie" entirely and just call it "dml".

matmarzolagomes (approved):

PR #112, opened 2025-11-13
feat: add dml parser. 

i4ki:


dml/parser_test.go
```diff
@@ -0,0 +1,253 @@
...
+					},
+				},
+			},
+		},
+		{
+			text: `SET feedbacks .={} WHERE id    =   1;`,
+			want: dml.Stmts{
+				{
+					Op:     dml.SET,
+					Entity: u("feedbacks"),
```
katcipis: hmm do you believe the unique thing will make a big difference here ? I only saw it on the release note of some recent Go version...and it reminded me of Lua since it has internalized strings by default/all the time.

katcipis (approved):

PR #115, opened 2025-11-26
feat: add xjson.DynSet

katcipis:


xjson/xjson.go
```diff
@@ -140,10 +145,54 @@ func DynGet[T any](o Obj, path string) (T, error) {
...
+//
+// Path is defined using '.' as delimiter like: "key.nested1.nested2.nested3".
+//
+// Key names with "." can be traversed by using double quotes like:
+//   - "key."nested.dot".key2"
+//
+// It will traverse key -> nested.dot -> key2 and set "key2" to be the given value.
+// If the given path is invalid, like "" or "." or the [Obj] is nil an error is returned.
+func DynSet(o Obj, path string, value any) error {
+	if o == nil {
```
i4ki: this is the kind of place where this is reasonable IMO
in search I removed lots of defensive `if`s in the provided arguments...
but when you have something that potentially can be used to recursively traverse stuff, then it makes a lot of sense.

PR #121, opened 2025-12-04
feat: add support for '-' in identifier and quoted path in traversal. 

i4ki:
- '-' in identifiers

Now it's possible to use: 
```
SET feedbacks-2025-01-01 custom_fields.my-dashed-field = "test" WHERE my-id = 1;
```
Usually our core fields will not contain dashes but this is needed for user-defined fields.

Note: '-' cannot be a leading or trailing character.

- Quoted path traversal
```
SET feedbacks custom_fields."this is a crazy field" = 1 WHERE id=1;
```

### DIFF
--- a/dml/token.go
+++ b/dml/token.go
@@ -5,7 +5,7 @@ import (
 	"unicode/utf8"
 )
 
-// identifier: \p{L}+[\p{L}0-9_-]*[\p{L}0-9]+$ with support for matched quotes
+// identifier: ^\p{L}(?:[\p{L}0-9_-]*[\p{L}0-9])?$
 func isIdent(s string) bool {
 	data := []byte(s)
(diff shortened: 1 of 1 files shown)


dml/token.go
```diff
@@ -5,28 +5,24 @@ import (
 	"unicode/utf8"
 )
 
-// identifier: \p{L}+[_\p{L}0-9]* with support for matched quotes
+// identifier: \p{L}+[\p{L}0-9_-]*[\p{L}0-9]+$ with support for matched quotes
```
gustavom2998: Maybe a dumb question, but wouldn't this be simpler for interpretation? Not sure if the validated content changes

```suggestion
// identifier: \p{L}+[\p{L}0-9_-]*[\p{L}0-9]?$ with support for matched quotes
```

PR #154, opened 2026-04-28
feat: add subscription option for prefetching events.

i4ki:
Adds an option for prefetch'ng events in both `OrderedGoogleSub` and `GoogleExperimentalBatchSubcription`.

I tried to not break anyone by using an "extra settings option" varArgs... I think an option in the constructor is better than a method because who knows if it could be changed once the subscription goroutines started.

I decided for an actual "size" instead of a multiplier because it seems more generic, as someone could always do:
```
WithSubscriptionPrefetch(myBatchSize)
WithSubscriptionPrefetch(myBatchSize / 2)
WithSubscriptionPrefetch(myBatchSize * 1.5)
etc
```
but the multiplier would never allow for a fixed value increase (like just prefetch +50 events, independent of the batch size).


event/google.go
```diff
@@ -24,6 +24,8 @@ type (
 		eventName string
 		client    *pubsub.Client
 		sub       *pubsub.Subscription
+
+		extraSettings subscriptionSettings
```
katcipis:
```suggestion
		opts subscriptionOptions
```

Mostly because it is smaller and I think it is more usual to call "option" in Go....but feel free to disagree/ignore.

event/google.go
```diff
@@ -40,7 +42,16 @@ type (
...
 	}
+
+	subscriptionSettings struct {
+		prefetch int
+	}
+
+	// SubscriptionSettingOption is an option for setting up the subscription settings.
+	SubscriptionSettingOption func(settings *subscriptionSettings)
```
katcipis: I dont know..seems like we are leaking something here...the "subscriptionSettins" is something internal..and it is a little confusing that we have an option to configure a setting...I was expecting just "Subscription now has some options" and that is it...but maybe Im missing something here 🤔

event/google.go
```diff
@@ -120,7 +131,7 @@ func (p *OrderedGooglePublisher[T]) Shutdown(context.Context) error {
 // similar to [NewSubscription]. Ordering affects how concurrency is handled. Concurrency is done by handling
 // different ordering keys/partitions, every ordered key will be handled sequentially only different ordering keys will be handled concurrently.
 // Call [OrderedGoogleSub.Shutdown] to stop all goroutines/clean up all resources.
-func NewOrderedGoogleSub[T any](ctx context.Context, project, subName, eventName string, maxConcurrentEvents int) (*OrderedGoogleSub[T], error) {
+func NewOrderedGoogleSub[T any](ctx context.Context, project, subName, eventName string, maxConcurrentEvents int, settingOptions ...SubscriptionSettingOption) (*OrderedGoogleSub[T], error) {
```
katcipis:
```suggestion
func NewOrderedGoogleSub[T any](ctx context.Context, project, subName, eventName string, maxConcurrentEvents int, opts ...SubscriptionOption) (*OrderedGoogleSub[T], error) {
```

Just thinking 🤔

event/google.go
```diff
@@ -129,8 +140,12 @@ func NewOrderedGoogleSub[T any](ctx context.Context, project, subName, eventName
 		return nil, fmt.Errorf("creating client: %w", err)
 	}
 	sub := client.Subscription(subName)
-	sub.ReceiveSettings.MaxOutstandingMessages = maxConcurrentEvents
-	return &OrderedGoogleSub[T]{eventName: eventName, client: client, sub: sub}, nil
+	orderedSub := &OrderedGoogleSub[T]{eventName: eventName, client: client, sub: sub}
+	for _, settingOption := range settingOptions {
+		settingOption(&orderedSub.extraSettings)
```
katcipis: I get it that this allow us to make the options "private"...but then introducing the whole "setting option" is confusing IMHO...having a separate internal struct only makes the option private (which is nice), although maybe just letting the option operate directly on OrderedGoogleSub would be fine for now (depends on how this evolves).

event/google.go
```diff
@@ -221,10 +236,23 @@ func NewGoogleExperimentalBatchSubscription[T any](ctx context.Context, project,
 		batchSize:     batchSize,
 		numGoroutines: numGoroutines,
 	}
+	for _, setting := range settings {
+		setting(&s.extraSettings)
```
katcipis: Ah you use on both types, makes perfect sense to have a private separate type then...but I would consider not leaking that, we could call this just SubscriptionOption and deal with it as an "option"...not a "setting option"...which is odd IMHO (I get the internal design...but externalizing this seems off..unless Im missing something).

event/google.go
```diff
@@ -40,7 +42,16 @@ type (
 		events        chan *Event[T]
 		batchSize     int
 		numGoroutines int
+
+		extraSettings subscriptionSettings
```
katcipis:
```suggestion
		opts subscriptionOptions
```

Just visualizing how it could be

event/google.go
```diff
@@ -40,7 +42,16 @@ type (
...
 	}
+
+	subscriptionSettings struct {
+		prefetch int
+	}
```
katcipis:
```suggestion
	subscriptionOptions struct {
		prefetch int
	}
```

katcipis (dismissed):
Just some bullshittery around naming/concepts, but feel free to disregard.

katcipis:
Forgot to say, I loved the options design/constructor only/immutable, great stuff :-)

event/google.go
katcipis:
```suggestion
	SubscriptionOption func(opts *subscriptionOptions)
```

Again...just visualizing

event/google.go
katcipis: hmm also I dont think we need to keep the extraSettings inside the instance..it is used only on the constructor to define MaxOutstandingMessages.

i4ki: the idea of using a separate type for the "Option" is because the same option is reused between `OrderedGoogleSub` and `ExperimentalBatchSub`.

i4ki: https://github.com/birdie-ai/golibs/pull/154/changes#r3155968685
haha saw this now

katcipis: my bad... I comment on the review as I read like an idiot :laughing:

event/google.go
```diff
@@ -129,8 +140,12 @@ func NewOrderedGoogleSub[T any](ctx context.Context, project, subName, eventName
...
 	}
 	sub := client.Subscription(subName)
-	sub.ReceiveSettings.MaxOutstandingMessages = maxConcurrentEvents
-	return &OrderedGoogleSub[T]{eventName: eventName, client: client, sub: sub}, nil
+	orderedSub := &OrderedGoogleSub[T]{eventName: eventName, client: client, sub: sub}
+	for _, opt := range opts {
+		opt(&orderedSub.opts)
+	}
+	orderedSub.sub.ReceiveSettings.MaxOutstandingMessages = maxConcurrentEvents + orderedSub.opts.prefetch
+	return orderedSub, nil
```
katcipis, line 148:
```suggestion
	orderedSub := &OrderedGoogleSub[T]{eventName: eventName, client: client, sub: sub}
	o := subscriptionOptions{}
	for _, opt := range opts {
		opt(&o)
	}
	orderedSub.sub.ReceiveSettings.MaxOutstandingMessages = maxConcurrentEvents + o.prefetch
	return orderedSub, nil
```

Rationale: If we never use the options other than in the constructor making this explicit/clear is an advantage IMHO (we can be sure it is used only here + less fields on the OrderedGoogleSub). Maybe I missed something and it is used somewhere else, disregard in that case.

katcipis (approved):

PR #158, opened 2026-05-27
feat: DML pruning by stmt

matmarzolagomes:
Implementing "PRUNING BY" syntax.

Example:
```
SET feedbacks
  custom_fields.abc={ ... }
WHERE id="abc"
PRUNING BY
  posted_at="2026-05-27 15:04:28+00:00",
  ingested_at="2026-05-28 15:04:28+00:00";
```

Pruning should be used as a "tip" to identify partitions and improve some SET stmts.

### DIFF
--- a/dml/parser.go
+++ b/dml/parser.go
@@ -122,6 +122,11 @@ func parseStmtExpr(in []byte) (Stmt, []byte, error) {
 	if err != nil {
 		return Stmt{}, nil, err
 	}
+	in = skipblank(in)
+	stmt.Pruning, in, err = parsePruning(in)
+	if err != nil {
+		return Stmt{}, nil, err
+	}
 	return stmt, in, nil
 }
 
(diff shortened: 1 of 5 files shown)


dml/parser.go
```diff
@@ -122,6 +122,11 @@ func parseStmtExpr(in []byte) (Stmt, []byte, error) {
 	if err != nil {
 		return Stmt{}, nil, err
 	}
+	in = skipblank(in)
+	stmt.Pruning, in, err = parsePruning(in)
+	if err != nil {
+		return Stmt{}, nil, err
+	}
```
i4ki, line 150: parsePruning must be called *if* a ` PRUNING BY ` is found.

```suggestion
	ident, in, err = lexIdent(in)
	isNotIdent := errors.Is(err, ErrNotIdent)
	if err != nil && !isNotIdent {
		return Stmt{}, nil, err
	}
	if isNotIdent {
	    return stmt, in, nil
	}
	if !strings.EqualFold(ident, "PRUNING") {
		return Stmt{}, nil, fmt.Errorf("%w: unexpected token %q", ErrSyntax, ident)
	}
	ident, in, err = lexIdent(in)
	if err != nil {
		return Stmt{}, nil, fmt.Errorf(`%w: expected BY keyword`, ErrSyntax, err)
	}
	if !strings.EqualFold(ident, "BY") {
		return Stmt{}, nil, fmt.Errorf("%w: unexpected token %q", ErrSyntax, ident)
	}
	stmt.Pruning, in, err = parsePruning(in)
	if err != nil {
		return Stmt{}, nil, err
	}
```
and then remove the `len(in) < 7 || !bytes.EqualFold(in[:7], []byte("PRUNING"))` check from inside the `parsePruning`, as it silently skips over malformed user input and postpone the error to the final `semicolon` check, which can be misleading as the "stmt parser func" does not fully parse a statement.

i4ki (dismissed):
I left an improvement idea but if you are in a hurry then go ahead and merge if it's working.

Thank you!

dml/parser.go
matmarzolagomes: i was testing and found a issue here, lexIdent() returns nil when it errors and when its not ident, it will fail when checking the last token ";".

Two options:
- save "in" on another variable and return it when isNotIdent:
```
saved := in
	ident, in, err = lexIdent(in)
	isNotIdent := errors.Is(err, ErrNotIdent)
	if err != nil && !isNotIdent {
		return Stmt{}, nil, fmt.Errorf("%w: %w", ErrSyntax, err)
	}
	if isNotIdent {
		return stmt, saved, nil
	}
```

Or, second option, change lexIdent to return "in" when it error, but that could sound weird... its returning a error, so I would expect to get the zero value, which is nil.

In the current change i went with the first option, any suggestions? or go back to checking len(in) < 7?

i4ki (approved):
LGTM

